PR #35, opened 2018-10-15
Save token rate to db

halink0803:


tradelogs/storage/influx.go
```diff
@@ -188,3 +189,48 @@ func (is *InfluxStorage) tradeLogToPoint(log common.TradeLog) ([]*client.Point,
...
+		err error
+	)
+	tags := map[string]string{
+		"block_number": strconv.FormatUint(rate.BlockNumber, 10),
+	}
+	fields := map[string]interface{}{
+		"provider": rate.Provider,
+		"rate":     rate.Rate,
+	}
+	pt, err = client.NewPoint("tokenrate", tags, fields, rate.Timestamp)
```
namnm1991: I think token_rate is better name, easier to read.

tradelogs/cmd/trade-logs-crawler/main.go
```diff
@@ -122,7 +122,7 @@ func getTradeLogs(c *cli.Context) error {
 		return err
 	}
 
-	tradeLogs, err := crawler.GetTradeLogs(fromBlock, toBlock, time.Second*5)
+	tradeLogs, usdRates, err := crawler.GetTradeLogs(fromBlock, toBlock, time.Second*5)
```
favadi: it does not make sense to me to have the crawler.GetTradeLogs returned the usdRates. I think we could move the fetch rates to a separate lib. Maybe we could have something like `ETHUSDRateFetcher` struct that have:

- influxdb client
- database name

and store the database there.

PR #40, opened 2018-10-17
Save eth usd rate to trades measurement

halink0803:


favadi:
[Eth-Usd historical rate](https://trello.com/c/sTM5PPzI/274-eth-usd-historical-rate)

tradelogs/cmd/trade-logs-crawler/main.go
```diff
@@ -144,21 +145,27 @@ func getTradeLogs(c *cli.Context) error {
 		return err
 	}
 
-	err = influxStorage.SaveTradeLogs(tradeLogs)
-	if err != nil {
-		return err
-	}
-
 	// fetch eth usd rate
 	ethUSDRateFetcher, err := tokenrate.NewETHUSDRateFetcher(sugar, dbName, influxClient, coingecko.New())
```
cyberliem: Should this be cache provider. Following this logic then it will always be calling to coingecko's API for ETH-USD rate.

tradelogs/storage/influx.go
```diff
@@ -127,6 +130,7 @@ func (is *InfluxStorage) tradeLogToPoint(log common.TradeLog) ([]*client.Point,
 		"src_amount":  srcAmount,
 		"dst_amount":  dstAmount,
 		"fiat_amount": log.FiatAmount,
+		"rate":        rate.Rate,
```
cyberliem: I would suggest ETHUSD_rate for this field name, otherwise it's really ambiguous.

tradelogs/cmd/trade-logs-crawler/main.go
halink0803:
```suggestion
	ethUSDRateFetcher, err := tokenrate.NewETHUSDRateFetcher(sugar, dbName, influxClient, coingecko.New())
```
As @favadi said, the provider might change and it could be different about the precision (now is by day but it could be hour or minute or even second), so that we leave it to call one time per block.

PR #43, opened 2018-10-18
Trade logs api review

favadi:


favadi:
[Implement Trade Events components](https://trello.com/c/osu3Cnlt/256-implement-trade-events-components)

PR #49, opened 2018-10-23
move database name to common package

favadi:


favadi:
[Implement users component](https://trello.com/c/Ras0ZhqQ/253-implement-users-component)

PR #54, opened 2018-10-24
Drop test db after finish testing trade log storage

namnm1991:


namnm1991:
[Teardown() in influx_result_test](https://trello.com/c/R93OQNoB/299-teardown-in-influxresulttest)

PR #55, opened 2018-10-24
delete conditions that could not happen

cyberliem:


cyberliem:
[Unlogical condition fix](https://trello.com/c/88DyTkaE/302-unlogical-condition-fix)

PR #56, opened 2018-10-24
Trade logs crawler concurrent

favadi:


favadi:
[Implement Trade Events components](https://trello.com/c/osu3Cnlt/256-implement-trade-events-components)

tradelogs/workers/workers.go
```diff
@@ -0,0 +1,215 @@
...
+					"to", to.String())
+
+				// Compare the job order to lastCompletedOrder and save to database if job order = lastCompletedOrder + 1,
+				// otherwise blocks worker
+				saveSuccess := false
+				for {
+					var err error
+
+					p.mutex.Lock()
+					if order == p.LastCompletedJobOrder+1 {
```
cyberliem: This is prone to deadlock when a worker met an error half way through when the other is waiting for it result. 

To reproduce the error, run with the following flags : --from-block=6494000   --to-block=6494500 --max-workers=3 --max-blocks=100.

At the worker's code (line 140), inject this condition to make it failed everytime from block is 6494300:
```Go
	if err != nil || (from.Int64() == 6494300) {
```
So the worker failed after the first round of round (3 rounds succeeded, but the 4th failed). This will hang the program. 

I suggest to use errgroup, similar to [reserveRate](https://github.com/KyberNetwork/reserve-stats/blob/561cd6ca739a1fa6d61955e033b726a83d6577ea/reserverates/crawler/crawler.go#L125), ensure the code exit if one of the thread failed.

namnm1991: Thanks @cyberliem, nice catch. I fix it here 8cbd2a9ba220228dec04adba12b66a2ab35a5802

tradelogs/cmd/trade-logs-crawler/main.go
```diff
@@ -204,12 +204,18 @@ func run(c *cli.Context) error {
 		sugar.Debugw("number of fetcher jobs", "jobs", jobs, "max_blocks", maxBlock)
 
 		go func(fromBlock, toBlock, maxBlocks int64) {
-			var jobOrder = p.LastCompletedJobOrder
+			var jobOrder = p.GetLastCompleteJobOrder()
 			for i := int64(fromBlock); i < toBlock; i = i + maxBlocks {
 				jobOrder++
 				p.Run(workers.NewFetcherJob(c, jobOrder, big.NewInt(i), big.NewInt(i+maxBlocks)))
```
cyberliem: This should be 
```go
  p.Run(workers.NewFetcherJob(c, jobOrder, big.NewInt(i), big.NewInt(min(i+maxBlock, toBlock))))
```

tradelogs/storage/influx.go
```diff
@@ -62,9 +63,33 @@ func (is *InfluxStorage) SaveTradeLogs(logs []common.TradeLog, rates []tokenrate
...
+		return 0, err
+	}
+
+	if len(res) != 1 || len(res[0].Series) != 1 || len(res[0].Series[0].Values[0]) != 3 {
+		is.sugar.Info("no result returned for last block query")
+		return 0, nil
+	}
 
-	return nil
+	return influxdb.GetInt64FromInterface(res[0].Series[0].Values[0][1])
```
cyberliem: When running with no lastBlock, this function return error 
```2018/10/30 11:28:52 invalid int64 value 6494166```

Comment (namnm1991) [reply tradelogs/storage/influx.go]: fixed in 92a701d251464227e3abb07937256a0adb5ee6bb

File: tradelogs/cmd/trade-logs-crawler/main.go (hunk 2 above)
Comment (namnm1991) [reply tradelogs/cmd/trade-logs-crawler/main.go]: changed in 5dddd3940668e864dd285e0b66c55da5a98228f8

File: tradelogs/workers/workers.go [hunk 4]
```diff
@@ -0,0 +1,224 @@
...
+	"go.uber.org/zap"
+)
+
+const dbName = "trade_logs"
+
+type executeJob func(*zap.SugaredLogger) ([]common.TradeLog, error)
+
+type job interface {
+	// retry the given fn function for attempts time with sleep duration between before returns an error.
+	retry(fn executeJob, attempts int, logger *zap.SugaredLogger) ([]common.TradeLog, error)
```
favadi: I don't feel like retry should be a method of job interface. It could be a stand alone function as I see.

favadi: fixed that in c23f2409681b4889fa85ed9aeee923b5cc0ca36a.

PR #58, opened 2018-10-24
Using coversion functions from lib and improve test

halink0803:


halink0803:
[Use lib/influxdb/conversion utils in reserverates](https://trello.com/c/PuOX5w0F/297-use-lib-influxdb-conversion-utils-in-reserverates)

PR #61, opened 2018-10-25
add export.sh script for automatically export tradelogs sample data

favadi:


favadi:
[Improve tradelogs result test](https://trello.com/c/u8u68xmG/303-improve-tradelogs-result-test)

tradelogs/storage/testdata/export.sh
```diff
@@ -41,8 +41,9 @@ docker exec "$container_name" influx_inspect export -datadir /var/lib/influxdb/d
     -waldir /var/lib/influxdb/wal -out /var/lib/influxdb/data/export.dat \
     -database trade_logs -retention autogen
 
+docker exec "$container_name" sed -i -e '/^#\|^CREATE DATABASE/d' /var/lib/influxdb/data/export.dat 
+
```
cyberliem: Should we sed after the copy version? Original version need not to be changed, it might be useful for other import.

favadi: we do not keep the unaltered version, as we cleanup everything in L49. Also, we can always rerun the script if need.

PR #68, opened 2018-10-30
Burn fee event

namnm1991:


namnm1991:
[Burn fee events](https://trello.com/c/0rmbTYek/309-burn-fee-events)

lib/blockchain/utils.go
```diff
@@ -22,3 +26,14 @@ func floatToBigInt(amount float64, decimal int64) *big.Int {
 func EthToWei(n float64) *big.Int {
 	return floatToBigInt(n, 18)
 }
+
+// IsBurnable indicate if the burn fee event was emitted when
+// the given token was trade on KyberNetwork
+func IsBurnable(t ethereum.Address) bool {
+	for _, token := range notBurnTokens {
```
favadi: please turn `notBurnTokens` to a `map[common.Address]struct{}` so we don't have to scan all array here.

lib/blockchain/token_addresses.go
```diff
@@ -10,4 +10,10 @@ var (
 
 	// KNCAddr is KNC token address
 	KNCAddr = common.HexToAddress("0xdd974D5C2e2928deA5F71b9825b8b646686BD200")
+
+	// WETHAddr is wrapped ETH address
+	WETHAddr = common.HexToAddress("0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2")
+
+	// KCCAddr is Kyber
```
favadi: the comment is incomplete.

tradelogs/storage/influx.go
```diff
@@ -270,13 +270,18 @@ func (is *InfluxStorage) tradeLogToPoint(log common.TradeLog, rate tokenrate.ETH
...
+		reserveAddr := burn.ReserveAddress.String()
+
+		_, ok := tags["src_rsv_addr"]
+		if !ok && blockchain.IsBurnable(log.SrcAddress) {
+			// source reserve address was not set, and the source token
+			// has burn event
+			tags["src_rsv_addr"] = reserveAddr
+		} else {
+			tags["dst_rsv_addr"] = reserveAddr
+		}
```
favadi: Maybe those codes will do the samething, but I feel like something this maybe easier to debug later.

```go
if blockchain.IsBurnable(log.SrcAddress) {
  if blockchain.IsBurnable(log.DestAddress) {
  // assert if len Burn == 0
  } else {
    // assert if len == 1 and set dst_rsv_addr tag
  }
} else {
   if blockchain.IsBurnable(log.DestAddress) {
  // assert if len Burn == 1 and set src_svr_addr
  } else {
    // assert if len == 2 and and both tags
  }
}

lib/blockchain/token_addresses.go
namnm1991: 11e492a593b629b385f94fca3fe7acd22dccf69d

lib/blockchain/utils.go
namnm1991: 382a58c998597a1449aa5925260c1900fef301d4

tradelogs/storage/influx.go
namnm1991: 42ac5fc33ceaab8bbe6372489fae52984c02e665

PR #69, opened 2018-10-30
Multiple deployment

favadi:


favadi:
[Support multiple running modes](https://trello.com/c/mY8bnJLv/278-support-multiple-running-modes)

lib/app/address.go
```diff
@@ -0,0 +1,41 @@
...
+// NewCrossDeploymentAddress returns an Address with given same address for all deployments.
+func NewCrossDeploymentAddress(addr []common.Address) Address {
+	return map[deployment.Mode][]common.Address{
+		deployment.Production: addr,
+		deployment.Staging:    addr,
+	}
+}
+
+// MustGetFromContext returns the common address for given deployment from context.
+func (a Address) MustGetFromContext(c *cli.Context) []common.Address {
```
favadi: please add a `MustGetOneFromContext` or `MustGetSingleFromContext` that panic if len of addresses != 1 and use it instead of `[0]` everywhere in consumer.

lib/app/app.go
```diff
@@ -21,18 +35,27 @@ func NewApp() *cli.App {
 			Usage: "app running mode",
 			Value: developmentMode,
 		},
+		cli.IntFlag{
+			Name:  Flag,
+			Usage: "Kyber Network deployment name",
+			Value: int(deployment.Production),
```
favadi: this is not nice as user don't know what 0, 1, 2, 3 means. It should accept string "production" (default), "staging" and convert it to `Deployment` type somewhere.

PR #70, opened 2018-10-31
fix the http proxy tests

favadi:


favadi:
[Implement reserve stats API gateway](https://trello.com/c/rC33h8yo/257-implement-reserve-stats-api-gateway)

PR #71, opened 2018-10-31
add old contract addresses to trade log crawler

favadi:


favadi:
[Missing addresses in trade logs crawler](https://trello.com/c/RxJAcAVt/310-missing-addresses-in-trade-logs-crawler)

PR #72, opened 2018-10-31
Gateway httpsign

ductm54:


lib/httputil/httprequestsign.go
```diff
@@ -0,0 +1,100 @@
...
+)
+
+const (
+	requestTarget = "(request-target)"
+	date          = "date"
+	digest        = "digest"
+	host          = "host"
+)
+
+//TODO: add signing algorithm
```
favadi: still relevant? IMO, we can only supports HMAC512 for now in our application.

lib/httputil/httprequestsign.go
```diff
@@ -0,0 +1,100 @@
...
+
+//Sign sign a http request with selected header and return a signed request
+//Use with http client only
+func Sign(r *http.Request, headers []string, keyID string, secret string) (*http.Request, error) {
```
favadi: `key ID string, secret string` can be shortened to `keyID, secret string`.

lib/httputil/httprequestsign.go
```diff
@@ -0,0 +1,100 @@
...
+	if len(headers) == 0 {
+		headers = []string{"date"}
```
favadi: replace "data" with the defined const.

PR #74, opened 2018-11-01
add cq into asset volume

cyberliem:


lib/cq/cq.go
```diff
@@ -102,10 +102,12 @@ func (cq *ContinuousQuery) prepareQueries(isCQ bool) ([]string, error) {
 		}
 		// Modify name and query so that mutiple queries are allowed
 		actualName := cq.Name
+		actualQuery := cq.Query
 		if offsetInterval != "" {
+			// return nil, fmt.Errorf("fuck this shis |%s|", offsetInterval)
 			actualName = actualName + "_" + offsetInterval
+			actualQuery = modifyINTOclause(cq.Query, offsetInterval)
```
favadi: so this is a bug that sneaks out of the test suite, we should have a task for it later.

tradelogs/cmd/trade-logs-crawler/main.go
```diff
@@ -248,6 +252,28 @@ func run(c *cli.Context) error {
 			break
 		}
 	}
-
+	//Deploy CQ	after get/ store tradelog
+	deploy := c.Bool(cq.CqsDeployFlag)
+	execute := c.Bool(cq.CqsExecuteFlag)
+	if deploy || execute {
```
favadi: those codes will never executed if crawler in daemon mode, it should be move up before fetching trade logs.

tradelogs/cmd/trade-logs-crawler/main.go
```diff
@@ -248,6 +252,28 @@ func run(c *cli.Context) error {
...
+		cqs, err := tradelogcq.CreateAssetVolumeCqs(common.DatabaseName)
+		if err != nil {
+			return err
+		}
+		if deploy {
+			for _, cq := range cqs {
```
favadi: cq conflicts with package name cq.

tradelogs/cmd/trade-logs-crawler/main.go
```diff
@@ -248,6 +252,28 @@ func run(c *cli.Context) error {
...
+			}
+		}
+		if execute {
+			for _, cq := range cqs {
+				if dErr := cq.Execute(influxClient, sugar); dErr != nil {
+					return dErr
+				}
+			}
+		}
+	}
```
favadi: this codes could be wrap in a function to be reused in both trade-logs-crawler and reserve-rates-crawler. Something like

```go
// in lib/cq/flags.go

func ManageCQFromContext(c *cli.Context, []*cq.ContinuousQuery) error {
    if deploy { ..deploy};
    if execute(...execute; exit 0};
}
```

favadi: And this function should be placed before the actual log fetching is done.

favadi:
the query in .sql/readme.md should be removed as they are in codes now.

lib/cq/cq.go
cyberliem: task created. 
https://trello.com/c/KNf3b1u2/315-add-test-case-when-offsetinterval-for-cq

PR #77, opened 2018-11-04
Reserve rates concurrent

namnm1991:


namnm1991:
[Implement concurrent workers pool for reserve rates crawler](https://trello.com/c/ir3cI7Ne/313-implement-concurrent-workers-pool-for-reserve-rates-crawler)

PR #78, opened 2018-11-05
Exit immediately when the workers pool encounterd error

namnm1991:


namnm1991:
[TradeLog Crawler Infinity loop](https://trello.com/c/XuZI7V0k/319-tradelog-crawler-infinity-loop)

PR #79, opened 2018-11-05
Add gateway dockerfile and build config

vietthang207:


vietthang207:
[Write roles for reserve-stats](https://trello.com/c/qVdhdgjs/135-write-roles-for-reserve-stats)

PR #82, opened 2018-11-06
Trade summary aggregation

cyberliem:


lib/cq/cq.go
```diff
@@ -86,6 +86,28 @@ func modifyINTOclause(query string, suffix string) string {
 	return strings.TrimSuffix(s, " ")
 }
 
+// hasGroupBy decide if the query has a GROUP BY from the main select
+// the logic is to count the number of complete bracket from the last GROUP BY
+func hasGroupBy(query string) bool {
```
favadi: I added tests for this in dd3516eb3cd2329c3c55bad6585662a47a46951b to prevent regression.

tradelogs/http/trade_summary.go
```diff
@@ -2,18 +2,38 @@ package http
...
+	addressesTableName = "addresses"
+)
+
 type tradeSummaryQuery struct {
 	From     uint64 `form:"fromTime"`
 	To       uint64 `form:"toTime"`
 	Timezone uint64 `form:"timezone" binding:"required,isValidTimezone"`
 }
 
+func (sv *Server) countKYCEDAddresses(ts uint64) (uint64, error) {
```
cyberliem: I think this function should be put into userDB interface. Server will only get the result from userDB, it shouldn't handle the logic 
@favadi  please advise if my speculation is correct.

favadi: Yes, it should be moved to db package, not http handler.

tradelogs/storage/trade_summary.go
```diff
@@ -0,0 +1,108 @@
...
+		return 0, nil, err
+	}
+	return tsUint64, &common.TradeSummary{
+		ETHVolume:       ethVolume,
+		USDAmount:       usdVolume,
+		TotalTrade:      totalTrade,
+		UniqueAddresses: uniqueAddr,
+		USDPerTrade:     usdPerTrade,
+		ETHPerTrade:     ethPerTrade,
+		NewUquAddresses: newUnqAddress,
```
favadi: NewUquAddresses  is bad abbreviation, should just use `NewUniqueAddresses`.

tradelogs/storage/cq/summary_cqs.go
```diff
@@ -0,0 +1,54 @@
...
+		"1d",
+		[]string{},
+	)
+	if err != nil {
+		return nil, err
+	}
+	result = append(result, volCqs)
+
+	newUnqAddressCq, err := libcq.NewContinuousQuery(
+		"summary_volume",
```
favadi: both this CQ and the on in L25 is "summary_volume", so this won't work.

tradelogs/storage/influx.go
```diff
@@ -339,5 +340,40 @@ func (is *InfluxStorage) tradeLogToPoint(log common.TradeLog) ([]*client.Point,
...
+				return nil, err
+			}
+			points = append(points, firstTradePt)
+		}
+	}
 	return points, nil
 }
+
+func (is InfluxStorage) userTraded(addr ethereum.Address) (bool, error) {
+	q := fmt.Sprintf("SELECT traded FROM first_trades WHERE user_addr='%s'", addr.String())
```
favadi: this is very wrong, we will vulnerable to SQL injection. We need to construct the query with proper parameter passing.

cyberliem: this function is not export, and require no user input. In addition to that, the parameter is needed to be an ethereum.Address. Hence there is no validation required for the param.

favadi: I thought that is PostgresQL query. It is true we couldn't do better with InfluxDB query, ignore it then.

PR #90, opened 2018-11-11
User volume

halink0803:


tradelogs/http/user_volume_query.go
```diff
@@ -0,0 +1,36 @@
...
+		return
+	}
+
+	_, _, err := query.Validate()
+	if err != nil {
+		httputil.ResponseFailure(c, http.StatusBadRequest, err)
+		return
+	}
+
+	volume, err := sv.storage.GetUserVolume(ethereum.HexToAddress(query.UserAddress), query.From, query.To, query.Freq)
```
favadi: please make this accept time.Time instead of unix millis.

favadi: still not done.

PR #92, opened 2018-11-12
User list api

halink0803:


gateway/http/http.go
```diff
@@ -72,6 +72,7 @@ func NewServer(addr, tradeLogsURL, reserveRatesURL, userURL, priceAnalyticURL,
 		r.GET("/asset-volume", tradeLogsProxyMW)
 		r.GET("/reserve-volume", tradeLogsProxyMW)
 		r.GET("/wallet-fee", tradeLogsProxyMW)
+		r.GET("/user-list", auth.Authenticated(), tradeLogsProxyMW)
```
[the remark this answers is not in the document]
favadi: please add document for this API.

tradelogs/http/user_list_query.go
```diff
@@ -0,0 +1,41 @@
...
+			err,
+		)
+		return
+	}
+	_, _, err := query.Validate()
+	if err != nil {
+		httputil.ResponseFailure(c, http.StatusBadRequest, err)
+		return
+	}
+	userList, err := sv.storage.GetUserList(query.From, query.To)
```
favadi: actually we have a task to change everything to accept time.Time instead of unix millis, please do that here.

tradelogs/storage/user_list.go
```diff
@@ -0,0 +1,56 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+	"log"
+
+	"github.com/KyberNetwork/reserve-stats/tradelogs/common"
+)
+
+const measurement = "user_volume_day"
```
favadi: please query trade_logs measurement directly/

tradelogs/storage/user_list.go
```diff
@@ -0,0 +1,56 @@
...
+
+//GetUserList return list of user info
+func (is *InfluxStorage) GetUserList(fromTime, toTime uint64) ([]common.UserInfo, error) {
+	var (
+		err error
+	)
+	result := []common.UserInfo{}
```
favadi: move this inside var group, and use nil slice declaration.

tradelogs/storage/user_list.go
```diff
@@ -0,0 +1,56 @@
...
+
+	logger := is.sugar.With("from time", fromTime, "to time", toTime)
```
favadi: we should add func name in logger context too.

tradelogs/storage/user_list.go
```diff
@@ -0,0 +1,56 @@
...
+	q := fmt.Sprintf(`
+		SELECT sum(eth_volume) as eth_amount, sum(usd_volume) as usd_amount from "%s"
+		WHERE time >= %d%s AND TIME <= %d%s GROUP BY user_addr
+	`, measurement, fromTime, timePrecision, toTime, timePrecision)
+
+	res, err := is.queryDB(is.influxClient, q)
+	if err != nil {
+		return result, err
+	}
+	jsonValue, _ := json.Marshal(res)
```
favadi: don't ignore marshal error.

gateway/http/http.go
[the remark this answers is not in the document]
halink0803: the /users is already taken by user module, so I keep using this endpoint as /user-list
I added document to api docs.

PR #95, opened 2018-11-14
1455 Multiple execute trade

namnm1991:


tradelogs/crawler.go
```diff
@@ -233,24 +209,42 @@ func (crawler *Crawler) GetTradeLogs(fromBlock, toBlock *big.Int, timeout time.D
...
-			if result, err = updateTradeLogs(result, logItem, ts); err != nil {
+			err := updateTradeLog(&tradeLog, logItem)
+			if err != nil {
 				return result, err
 			}
+
+			// a tradeEvent indicate that all tradeLog information was fulfilled
+			// we add the tradeLog into result and prepare to fulfill an other
+			// tradeLog
+			if topic.Hex() == tradeEvent {
```
favadi: please add a debug log here with some basic trade log information in context.

tradelogs/crawler.go
```diff
@@ -233,24 +209,42 @@ func (crawler *Crawler) GetTradeLogs(fromBlock, toBlock *big.Int, timeout time.D
...
+		if len(logItem.Topics) == 0 {
+			return result, errors.New("log item has no topic")
+		}
+
 		topic := logItem.Topics[0]
 		switch topic.Hex() {
 		case feeToWalletEvent, burnFeeEvent, etherReceivalEvent, tradeEvent:
-			// add logItem to result
-			if result, err = updateTradeLogs(result, logItem, ts); err != nil {
+			err := updateTradeLog(&tradeLog, logItem)
```
favadi: err is already declared above, also we should us `if xxx; err != nil {}` form.

tradelogs/crawler.go
favadi: is it clearer if we move this logic to another case of switch statement? `case tradeEvent`?

tradelogs/common/types.go
```diff
@@ -46,6 +48,8 @@ type TradeLog struct {
 
 	ETHUSDRate     float64 `json:"-"`
 	ETHUSDProvider string  `json:"-"`
+
+	Index uint `json:"index"`
```
favadi: please add a comment for this field, something like log index of in transactions events.

tradelogs/crawler.go
```diff
@@ -116,80 +116,56 @@ func logDataToBurnFeeParams(data []byte) (ethereum.Address, ethereum.Hash, error
...
-	if !updateLastLog {
-		tradeLog = common.TradeLog{
-			BlockNumber:     logItem.BlockNumber,
-			TransactionHash: logItem.TxHash,
-			Timestamp:       ts,
-		}
-	}
-
-	switch logItem.Topics[0].Hex() {
+func updateTradeLog(tradeLog *common.TradeLog, eventLog types.Log) error {
```
favadi: I think it should be clearer if we can refactor this to a function that accept `[]types.Log` and output `[]TradeLog, error`.

tradelogs/common/types.go
namnm1991: changed in 0ebe190eea0b12e76b479634fc5d0f2c2dc8e586

tradelogs/crawler.go
namnm1991: added new function in fa5a5d3a2453f656e96278816a655531d10f1bc5

tradelogs/crawler.go
namnm1991: changed in fa5a5d3a2453f656e96278816a655531d10f1bc5

tradelogs/crawler.go
namnm1991: changed in fa5a5d3a2453f656e96278816a655531d10f1bc5 with log and seperated logic

PR #108, opened 2018-11-19
1447 intergration app name mapping

cyberliem:


integration-app-names/app_names.go
```diff
@@ -0,0 +1,68 @@
...
+	mapping map[ethereum.Address]string
+}
+
+//NewMapAddrAppName return a new instance of MapAddrAppName
+func NewMapAddrAppName() *MapAddrAppName {
+	var (
+		temp = make(map[ethereum.Address]string)
+	)
+	return &MapAddrAppName{
+		mapping: temp,
```
favadi: I don't think the temp variable is useful here, please remove it.

integration-app-names/app_names.go
```diff
@@ -0,0 +1,68 @@
...
+		tmp = make(map[string]string)
+	)
+	data, err := ioutil.ReadFile(path)
+	if err != nil {
+		return err
+	}
+	err = json.Unmarshal(data, &tmp)
+	if err != nil {
+		return err
+	}
```
favadi: please use `os.Open` and feed the file descriptor to `json.NewDecode(fd).Decode()` instead of reading whole file into memory.

integration-app-names/app_names.go
```diff
@@ -0,0 +1,68 @@
...
+		temp = make(map[ethereum.Address]string)
+	)
+	return &MapAddrAppName{
+		mapping: temp,
+		mutex:   &sync.Mutex{},
+	}
+}
+
+//LoadFromFile add mapping addr-appname from a file
+func (apn *MapAddrAppName) LoadFromFile(path string) error {
```
favadi: this function should not be exposed to the interface as it will be called once. Please refactor it to an option of `NewMapAddrAppName` constructor.

```go
type Option function(*MapAddrAppName))

func WithDataFile(path string) Option {}

 func NewMapAddrAppName(options ...Option) *MapAddrAppName {}
```

integration-app-names/app_names.go
```diff
@@ -0,0 +1,68 @@
...
+	apn.mutex.Lock()
+	defer apn.mutex.Unlock()
+	apn.mapping[addr] = name
+	return nil
+}
+
+//GetAddrToAppName return a map of address to app name
+func (apn *MapAddrAppName) GetAddrToAppName() map[ethereum.Address]string {
+	apn.mutex.Lock()
+	defer apn.mutex.Unlock()
```
favadi: this will make any of its consumer turns into single threaded application. Please use sync.RWMutex.

integration-app-names/cmd/app-anames-api/main.go
```diff
@@ -0,0 +1,61 @@
+package main
```
favadi: `integration-app-names/cmd/app-anames-api/main.go` looks consistent.
Please rename it to either one below:

- integration-app-names/cmd/integration-app-anames-api/main.go
- app-names/cmd/app-anames-api/main.go

integration-app-names/cmd/app-anames-api/main.go
```diff
@@ -0,0 +1,61 @@
...
+	app.Name = "Addre to Intergration App Name"
+	app.Action = run
+	app.Version = "0.0.1"
+	app.Flags = append(app.Flags)
+	app.Flags = append(app.Flags, httputil.NewHTTPCliFlags(httputil.AddrToAppName)...)
+	app.Flags = append(app.Flags,
+		cli.StringFlag{
+			Name:   dataFilePathFlag,
+			Usage:  "file path to address to app name json",
+			EnvVar: "APPNAME_DATA_PATH",
```
favadi: please remove the APPNAME prefix for EnvVar, we do not use them anywhere.

integration-app-names/http/server.go
```diff
@@ -0,0 +1,84 @@
+package http
+
+import (
+	"net/http"
+
+	appNames "github.com/KyberNetwork/reserve-stats/integration-app-names"
+	"github.com/KyberNetwork/reserve-stats/lib/httputil"              // import custom validator functions
```
favadi: wrong comment.

integration-app-names/http/server.go
```diff
@@ -0,0 +1,84 @@
...
+			http.StatusInternalServerError,
+			err,
+		)
+		return
+	}
+	c.JSON(http.StatusOK, gin.H{q.Address: q.AppName})
+}
+
+func (sv *Server) register() {
+	sv.r.GET("/addr-to-appname", sv.getAddrToAppName)
```
favadi: `addr-to-appname` is too specific, just call them `appnames`, `application-names` is good enough because we will add more than address to this API later.

lib/appname/appname-client.go
```diff
@@ -0,0 +1,82 @@
+package appname
```
favadi: please rename `lib/appname/appname-client.go` --> `lib/appname/appname_client.go` for consistent sake.

lib/appname/appname-client.go
```diff
@@ -0,0 +1,82 @@
...
+	url := fmt.Sprintf("%s/%s",
+		strings.TrimRight(c.url, "/"),
+		strings.Trim(endpoint, "/"),
+	)
+	req, err := http.NewRequest(method, url, nil)
+	if err != nil {
+		return nil, err
+	}
+	req.Header.Add("Accept", "application/json")
+	req.Header.Add("Content-Type", "application/x-www-form-urlencoded")
```
favadi: I don't think we use `x-www-form-urlencoded` anywhere.

lib/appname/appname-client.go
```diff
@@ -0,0 +1,82 @@
...
+	req, err := c.newRequest(http.MethodGet, endpoint)
+	if err != nil {
+		return nil, err
+	}
+	rsp, err := c.client.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	c.sugar.Debugw("Get Addr To App Name", "response", rsp)
+	c.sugar.Debugw("Get Addr To App Name", "body", rsp.Body)
```
favadi: duplicated log message.

favadi: also log decoded body should be more useful.

lib/appname/appname-client.go
```diff
@@ -0,0 +1,82 @@
...
+
+	defer rsp.Body.Close()
+	c.sugar.Debugw("Get Addr To App Name", "response", rsp)
```
favadi: duplicated message.

lib/appname/appname-client.go
```diff
@@ -0,0 +1,82 @@
...
+// Client is the the real implementation of addr to integration app name  interface.
+type Client struct {
+	sugar  *zap.SugaredLogger
+	client *http.Client
+	url    string
+}
+
+func (c *Client) newRequest(method, endpoint string) (*http.Request, error) {
+	var (
+		logger = c.sugar.With("method", method, "endpoint", endpoint)
```
favadi: please add func name to context of this logger.

lib/appname/flags.go
```diff
@@ -0,0 +1,39 @@
...
+	"fmt"
+
+	"github.com/go-ozzo/ozzo-validation"
+	"github.com/go-ozzo/ozzo-validation/is"
+	"github.com/urfave/cli"
+	"go.uber.org/zap"
+)
+
+const (
+	appNameURLFlag = "app-name-url"
```
favadi: once decided the name for this application, remember to update it here.

lib/appname/flags.go
```diff
@@ -0,0 +1,39 @@
...
+)
+
+// NewCliFlags returns cli flags to configure a core client.
+func NewCliFlags() []cli.Flag {
+	return []cli.Flag{
+		cli.StringFlag{
+			Name:   appNameURLFlag,
+			Usage:  "Core API URL",
```
favadi: wrong usage.

lib/httputil/ports.go
```diff
@@ -21,4 +21,7 @@ const (
 
 	// PriceAnalytic is the port number of Price Analytics API
 	PriceAnalytic HTTPPort = 8006
+
+	//AddrToAppName is the port number for Integration Address to AppName API
+	AddrToAppName HTTPPort = 8007
```
favadi: please rename this to actual package name.

tradelogs/crawler.go
```diff
@@ -29,6 +29,9 @@ const (
 	tradeEvent = "0x1849bd6a030a1bca28b83437fd3de96f3d27a5d172fa7e9c78e7b61468928a39"
 	// etherReceivalEvent is the topic of event EtherReceival(address indexed sender, uint amount).
 	etherReceivalEvent = "0x75f33ed68675112c77094e7c5b073890598be1d23e27cd7f6907b4a7d98ac619"
+
+	kyberSwapAppName = "KyberSwap"
+	appNameJSONPath  = "app_name.json"
```
favadi: what is this variable?

tradelogs/crawler.go
```diff
@@ -29,6 +29,9 @@ const (
 	tradeEvent = "0x1849bd6a030a1bca28b83437fd3de96f3d27a5d172fa7e9c78e7b61468928a39"
 	// etherReceivalEvent is the topic of event EtherReceival(address indexed sender, uint amount).
 	etherReceivalEvent = "0x75f33ed68675112c77094e7c5b073890598be1d23e27cd7f6907b4a7d98ac619"
+
+	kyberSwapAppName = "KyberSwap"
```
favadi: should this be moved to the `lib/appname`?  we might need it for trade-logs-api displaying purpose.

tradelogs/crawler.go
```diff
@@ -151,6 +157,12 @@ func (crawler *Crawler) assembleTradeLogs(eventLogs []types.Log) ([]common.Trade
 				Index:          log.Index,
 			}
 			tradeLog.WalletFees = append(tradeLog.WalletFees, walletFee)
+			//if a tradelog has feeToWalletEvent, it is default to be unknown intergration app
+			tradeLog.IntegrationApp = "unknown"
+			//if Wallet Address < maxUint64, it is KyberSwap
+			if walletAddr.Big().Cmp(big.NewInt(0).Exp(big.NewInt(2), big.NewInt(128), nil)) == -1 {
```
favadi: while the comment indicate that it is maxUint64, the actual code is comparing to 2^128 - 1. What is the correct rule here?

tradelogs/crawler.go
```diff
@@ -151,6 +157,12 @@ func (crawler *Crawler) assembleTradeLogs(eventLogs []types.Log) ([]common.Trade
 				Index:          log.Index,
 			}
 			tradeLog.WalletFees = append(tradeLog.WalletFees, walletFee)
+			//if a tradelog has feeToWalletEvent, it is default to be unknown intergration app
+			tradeLog.IntegrationApp = "unknown"
```
favadi: `third_party` instead of `unknown` is clearer.

favadi: this should be moved to the post processing trade logs after it is assembled. also all of the logics to determine if it is KyberSWAP or not should be grouped into one function that accept a `common.TradeLog` and return true/false if it is KyberSWAP or not. The logic order in task description should be honoured.

tradelogs/crawler.go
```diff
@@ -230,7 +246,7 @@ func (crawler *Crawler) GetTradeLogs(fromBlock, toBlock *big.Int, timeout time.D
 		return nil, err
 	}
 
-	result, err := crawler.assembleTradeLogs(logs)
+	result, err = crawler.assembleTradeLogs(logs)
 	if err != nil {
```
favadi: please group 249-250 to one line.

tradelogs/storage/influx.go
```diff
@@ -111,17 +111,17 @@ func (is *InfluxStorage) LoadTradeLogs(from, to time.Time) ([]common.TradeLog, e
 
 		q = fmt.Sprintf(
 			`
-		SELECT %[1]s FROM burn_fees WHERE time >= '%[4]s' AND time <= '%[5]s' GROUP BY tx_hash, trade_log_index;
-		SELECT %[2]s FROM wallet_fees WHERE time >= '%[4]s' AND time <= '%[5]s' GROUP BY tx_hash, trade_log_index;
-		SELECT %[3]s FROM trades WHERE time >= '%[4]s' AND time <= '%[5]s' GROUP BY tx_hash, log_index;
+		SELECT %[1]s FROM burn_fees WHERE time >= '%[4]s' AND time <= '%[5]s' GROUP BY tx_hash, trade_log_index FILL(0);
```
favadi: not sure if I understand the changes here, how it is related to app name integration?

tradelogs/http/http.go
```diff
@@ -58,6 +60,24 @@ func (sv *Server) getTradeLogs(c *gin.Context) {
...
+	if err != nil {
+		libhttputil.ResponseFailure(
+			c,
+			http.StatusInternalServerError,
+			err,
+		)
+		return
+	}
+	for i, log := range tradeLogs {
+		if len(log.WalletFees) > 0 {
```
favadi: we should only lookup if tradeLogs.IntegrationApp name is "unknown", "third_party", right?

lib/appname/appname-client.go
cyberliem: sometime the reply might not be the valid format to decode. If it can be decoded right away then this log won't be necessary.

PR #113, opened 2018-11-21
1451 implement userprofile package

cyberliem:


lib/userprofile/address_heap.go
```diff
@@ -0,0 +1,39 @@
...
+
+// AddrHeap is a min-heap of AddrNode
+type AddrHeap []AddrNode
+
+// Len return the length of addrHeap
+func (a AddrHeap) Len() int {
+	return len(a)
+}
+
+//Less return bool indicate the comparion between 2 AddrNode by index
```
favadi: comparion --> comparison.

lib/userprofile/cached_user_profile.go
```diff
@@ -0,0 +1,75 @@
...
+	*Client
+	mu           *sync.RWMutex
+	cached       map[ethereum.Address]UserProfile
+	MaxCacheSize int64
+	cacheSize    int64
+	minHeap      *AddrHeap
+}
+
+// NewCachedClient creates a new User Profile cached client instance.
+func NewCachedClient(client *Client, maxcache int64) *CachedClient {
```
favadi: `maxcache` --> `maxCache`.

lib/userprofile/cached_user_profile.go
```diff
@@ -0,0 +1,75 @@
...
+		cacheSize:    0,
+		minHeap:      h,
+	}
+}
+
+// LookUpUserProfile will look for the UserProfile of input addr in cache first
+// If this fail then it will query from endpoint
+func (cc *CachedClient) LookUpUserProfile(addr ethereum.Address) (UserProfile, error) {
+	cc.mu.Lock()
+	defer cc.mu.Unlock()
```
favadi: this will make LookUpUserProfile single-threaded. We use `cc.mu.RLock ` to check if profile exists and `cc.mu.Lock` to update it later.

lib/userprofile/flags.go
```diff
@@ -0,0 +1,65 @@
...
+	"fmt"
+
+	"github.com/go-ozzo/ozzo-validation"
+	"github.com/go-ozzo/ozzo-validation/is"
+	"github.com/urfave/cli"
+	"go.uber.org/zap"
+)
+
+const (
+	userprofileURLFlag        = "user-url"
```
favadi: should be confusing with resever-rates's users service, please use `users-profile-url`.

lib/userprofile/flags.go
```diff
@@ -0,0 +1,65 @@
...
+	userprofileSigningKeyFlag = "user-signing-key"
+	maxUserCacheFlag          = "max-user-cache"
```
favadi: same comments as above, please use user-profile || userprofile to avoid confusion.

lib/userprofile/flags.go
```diff
@@ -0,0 +1,65 @@
...
+		cli.Int64Flag{
+			Name:   maxUserCacheFlag,
+			Usage:  "max Cache Size for user profile client. Default to 1000",
+			EnvVar: "MAX_USER_CACHE",
+			Value:  maxUserCacheDefault,
+		},
+		cli.StringFlag{
+			Name:   userprofileSigningKeyFlag,
+			Usage:  "user profile Signing Key",
+			EnvVar: "USER_SIGNING_KEY",
```
favadi: same comments as above, please use user-profile || userprofile in env  var.

lib/userprofile/user_profile_client.go
```diff
@@ -0,0 +1,119 @@
...
+	mac := hmac.New(sha512.New, []byte(c.signingKey))
+	if _, err := mac.Write([]byte(msg)); err != nil {
+		return "", err
+	}
+	return hex.EncodeToString(mac.Sum(nil)), nil
+}
+
+// generateNonce returns nonce header required to use Core API,
+// which is current timestamp in milliseconds.
+func generateNonce() string {
```
favadi: are those functions duplicated from the core client, could we make it as less duplicated as possible by moving them in to a shared lib? Maybe `lib/httputil/legacysign` or something, please make up a name.

tradelogs/cmd/trade-logs-api/main.go
```diff
@@ -68,18 +68,31 @@ func main() {
...
+
+		userClient, err := userprofile.NewClientFromContext(sugar, c)
+		if err != nil {
+			return err
+		}
 
 		api := http.NewServer(influxStorage, httputil.NewHTTPAddressFromContext(c),
-			sugar, coreCachedClient, options...)
+			sugar, coreCachedClient,
+			userprofile.NewCachedClientFromContext(userClient, c),
```
favadi: the userprofile integration should be optional just like third party app names integration. Please make sure we can run trade-logs-http-api without userprofile service.

lib/userprofile/user_profile_client.go
cyberliem: I created issue [1862](https://redmine.knstats.com/issues/1862) to do this later. It involve refactoring a bunch of codes outside the user-profile scope.

lib/userprofile/address_heap.go
cyberliem: fixed in https://github.com/KyberNetwork/reserve-stats/pull/113/commits/a3c5ba525b1093e7baceb0a88b7f2677fdc400e0

lib/userprofile/cached_user_profile.go
cyberliem: fixed in https://github.com/KyberNetwork/reserve-stats/pull/113/commits/a3c5ba525b1093e7baceb0a88b7f2677fdc400e0

lib/userprofile/flags.go
cyberliem: fixed in https://github.com/KyberNetwork/reserve-stats/pull/113/commits/a3c5ba525b1093e7baceb0a88b7f2677fdc400e0

lib/userprofile/flags.go
cyberliem: fixed in https://github.com/KyberNetwork/reserve-stats/pull/113/commits/a3c5ba525b1093e7baceb0a88b7f2677fdc400e0

lib/userprofile/cached_user_profile.go
cyberliem: fixed in https://github.com/KyberNetwork/reserve-stats/pull/113/commits/a3c5ba525b1093e7baceb0a88b7f2677fdc400e0

lib/userprofile/flags.go
cyberliem: fixed in https://github.com/KyberNetwork/reserve-stats/pull/113/commits/a3c5ba525b1093e7baceb0a88b7f2677fdc400e0

tradelogs/cmd/trade-logs-api/main.go
cyberliem: fixed in https://github.com/KyberNetwork/reserve-stats/pull/113/commits/a3c5ba525b1093e7baceb0a88b7f2677fdc400e0

lib/userprofile/cached_user_profile.go
```diff
@@ -0,0 +1,83 @@
...
+		"address", addr.Hex(),
+	)
+	p, ok := cc.LookUpCache(addr)
+	if ok {
+		logger.Debugw("cache hit")
+		return p, nil
+	}
+
+	logger.Debugf("cache missed. Lookup from API endpoint and caching... Current cache size : %d", cc.cacheSize)
+	p, err := cc.Client.LookUpUserProfile(addr)
```
favadi: - could you double check what is the value of `p`, `err` in case of no user found? 

- do we cache the empty  `UserProfile` value?
  - if yes
    - do we have to implement an expired time for `UserProfile`?

tradelogs/cmd/trade-logs-api/main.go
```diff
@@ -68,18 +68,34 @@ func main() {
...
+		cachedUserClient := userprofile.NewCachedClientFromContext(userClient, c)
+		if cachedUserClient != nil {
+			options = append(options, http.WithUserProfile(cachedUserClient))
+		}
 		api := http.NewServer(influxStorage, httputil.NewHTTPAddressFromContext(c),
-			sugar, coreCachedClient, options...)
+			sugar, coreCachedClient,
+			options...)
+		err = api.Start()
+		if err != nil {
```
favadi: L95-96 should be in 1 line.

lib/userprofile/cached_user_profile.go
cyberliem: in case no user found, the api will return empty. It is not an error, since there's alot of trade on Kybernetwork with a non-KYCed address. 

Consider this case: a non-KYCed address, namely 0xabc trade 100 times that day. If we don't store map[0xabc]nil, everytime we saw that tradelog, we'll make a call to API server. In my test experience, this case happen quite frequent. Hence, a nil record is needed to be cached. 

I think we should implement an expired time for UserProfile, since an unknown address might submit his KYC. I wonder how long should we allow this expired time?

favadi: > I wonder how long should we allow this expired time?

It is just for display purpose, so even a day is fine for me. Anw, we can leave this for later improvement.

PR #119, opened 2018-11-26
1708 Update kyc swap detection rules

halink0803:


tradelogs/common/types.go
```diff
@@ -154,6 +154,16 @@ type Heatmap struct {
 	TotalKYCUser         int64   `json:"total_kyc_user"`
 }
 
+func isKyberWallet(addr ethereum.Address) bool {
+	wallets := []string{"0x440bBd6a888a36DE6e2F6A25f65bc4e16874faa9", "0xEA1a7dE54a427342c8820185867cF49fc2f95d43"}
```
favadi: we should turn this into 

```go
	var wallets = map[ethereum.Address]struct{}{
		ethereum.HexToAddress("0x440bBd6a888a36DE6e2F6A25f65bc4e16874faa9"): {},
		ethereum.HexToAddress("0xEA1a7dE54a427342c8820185867cF49fc2f95d43"): {},
	}
```

so we don't have to scan through the map every time.

PR #121, opened 2018-11-27
1716 Improve reserve rate crawler addresses flag behavior

halink0803:


reserverates/cmd/reserve-rates-crawler/main.go
```diff
@@ -35,6 +36,10 @@ const (
 	durationFlag         = "duration"
 	shardDurationFlag    = "shard-duration"
 	defaultShardDuration = time.Hour * 24
+
+	modeFlag              = "mode"
+	developMode           = "develop"
```
favadi: looks strange, why do we need them?

reserverates/cmd/reserve-rates-crawler/main.go
```diff
@@ -35,6 +36,10 @@ const (
...
+	kyberReserveAddresses = "0x63825c174ab367968EC60f061753D3bbD36A0D8F"
```
favadi: should use `lib.app/NewAddress` as we need to support both production and staging environment.

reserverates/cmd/reserve-rates-crawler/main.go
```diff
@@ -164,6 +169,13 @@ func run(c *cli.Context) error {
 	delayTime := c.Duration(delayFlag)
 
 	addrs := c.StringSlice(addressesFlag)
+	if len(addrs) == 0 {
+		if c.String("mode") == "develop" {
```
favadi: could expose the mode from lib/app somehow.

reserverates/cmd/reserve-rates-crawler/main.go
```diff
@@ -170,10 +166,9 @@ func run(c *cli.Context) error {
 
 	addrs := c.StringSlice(addressesFlag)
 	if len(addrs) == 0 {
-		if c.String("mode") == "develop" {
-			addrs = []string{kyberReserveAddresses}
-		} else {
-			return errors.New("addresses flag cannot be empty")
+		addresses := contracts.InternalReserveAddress().MustGetFromContext(c)
```
favadi: should use `MustGetOneFromContext` as there is only one internal reserve. Also, could you add a log message stating that using internal reserve as user does not supply any address.

PR #134, opened 2018-12-06
1448 Update integration app

halink0803:


app-names/storage/app_name_storage.go
```diff
@@ -0,0 +1,312 @@
...
+		var count int
+		if err := tx.Get(&count, fmt.Sprintf("SELECT COUNT(*) FROM %s WHERE address='%s'", addressesTableName, addressStr)); err != nil {
+			return app, err
+		}
+		logger.Debug(count)
+		if count > 0 {
+			if err := tx.Commit(); err != nil {
+				return app, err
+			}
+			return app, errors.New("address already exists")
```
cyberliem: should make an export variable for this

app-names/storage/app_name_storage.go
```diff
@@ -0,0 +1,312 @@
...
+		qresult []AppQueryResult
+		result  common.AppObject
+	)
+	logger.Debug("get an app")
+	if err := adb.db.Select(&qresult, fmt.Sprintf(`SELECT app_name.id, app_name.name, addresses.address 
+	from app_name JOIN addresses ON app_name.id = addresses.app_name_id WHERE app_name.id = %d AND app_name.active IS TRUE`, appID)); err != nil {
+		return result, err
+	}
+	if len(qresult) == 0 {
+		return result, errors.New("app does not exist")
```
cyberliem: should make an export variable for this

app-names/http/server.go
```diff
@@ -2,68 +2,201 @@ package http
...
+		)
+		return
+	}
+
+	result, err := sv.db.GetAppAddresses(appID)
+	if err != nil {
+		if err.Error() == "app does not exist" {
+			httputil.ResponseFailure(
+				c,
+				http.StatusNotFound,
```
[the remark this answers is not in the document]
halink0803: @favadi can you answer this question?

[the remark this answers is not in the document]
favadi: 404 is standard REST API return code for resource not found. `StatusNoContent` indicates same meaning as `StatusOK`, just that there is nothing to return in body. So, no, `StatusNoContent` is wrong in this case.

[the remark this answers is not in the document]
halink0803: Ok, so I will keep it as StatusNotFound.

app-names/http/server.go
```diff
@@ -2,68 +2,201 @@ package http
...
-func (sv *Server) getAddrToAppName(c *gin.Context) {
+func (sv *Server) getApps(c *gin.Context) {
 	var (
 		logger = sv.sugar.With("func", "intergration-app-names/http/Server.getAddrToAppName")
 	)
-
 	logger.Debug("getting addr to App name")
-	result := sv.addrToAppName.GetAddrToAppName()
+	name := c.Query("name")
+	logger.Debug(name)
```
favadi: bad log statement, should be `logger.Debugw("got name parameter from query", "name", name)` or add the name param to the logger directly.

app-names/http/server.go
```diff
@@ -2,68 +2,201 @@ package http
...
+		return
+	}
+	c.JSON(
+		http.StatusOK,
+		apps,
+	)
+}
+
+func (sv *Server) getAddressFromAppID(c *gin.Context) {
+	appIDStr := c.Param("appID")
```
favadi: param should be `app_id`?

favadi: also, should we use ShouldBindQuery instead?

app-names/http/server.go
```diff
@@ -2,68 +2,201 @@ package http
...
+	if err != nil {
+		httputil.ResponseFailure(
+			c,
+			http.StatusBadRequest,
+			err,
+		)
+		return
+	}
+
+	result, err := sv.db.GetAppAddresses(appID)
```
favadi: the function name is GetAppAddresses while it returns both address and name.

app-names/http/server.go
```diff
@@ -2,68 +2,201 @@ package http
...
+		} else {
+			httputil.ResponseFailure(
+				c,
+				http.StatusInternalServerError,
+				err,
+			)
+		}
+		return
+	}
+	if q.ID != 0 {
```
favadi: what is the logic behind this?

app-names/http/server.go
```diff
@@ -2,68 +2,201 @@ package http
...
+		q      common.AppObject
+	)
+	logger.Debug("start update app")
+	appID, err := strconv.ParseInt(c.Param("appID"), 10, 64)
+	if err != nil || appID == 0 {
 		httputil.ResponseFailure(
 			c,
-			http.StatusInternalServerError,
+			http.StatusBadRequest,
+			err,
```
favadi: if `appID == 0` what will returned in the error message?

app-names/http/server.go
```diff
@@ -2,68 +2,201 @@ package http
...
+		return
+	}
+	c.JSON(http.StatusOK, app)
+}
+
+func (sv *Server) deleteApp(c *gin.Context) {
+	var (
+		logger = sv.sugar.With("func", "app-names/server.deleteApp")
+	)
+	logger.Debug("delete app")
```
favadi: at least add query params to the context of the log message.

app-names/http/server.go
```diff
@@ -2,68 +2,201 @@ package http
...
+	appID, err := strconv.ParseInt(c.Param("appID"), 10, 64)
```
favadi: should we use ShouldBindQuery instead of ParseInt everywhere?

app-names/http/server.go
```diff
@@ -2,68 +2,201 @@ package http
...
 }
 
 func (sv *Server) register() {
-	sv.r.GET("/application-names", sv.getAddrToAppName)
-	sv.r.POST("/application-names", sv.updateAddrToAppName)
+	sv.r.GET("/application-names", sv.getApps)
+	sv.r.GET("/application-names/:appID", sv.getAddressFromAppID)
+	sv.r.POST("/application-names", sv.createApp)
+	sv.r.PUT("/application-names/:appID", sv.updateApp)
+	sv.r.DELETE("/application-names/:appID", sv.deleteApp)
```
favadi: do you think just `/applications` is better?

app-names/http/server_test.go
```diff
@@ -0,0 +1,344 @@
...
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/KyberNetwork/reserve-stats/app-names/storage"
+	"github.com/KyberNetwork/reserve-stats/lib/httputil"
+	"github.com/jmoiron/sqlx"
+	_ "github.com/lib/pq" // sql driver name: "postgres"
+	"github.com/stretchr/testify/assert"
+	"go.uber.org/zap"
```
favadi: please group and sort import statement.

app-names/http/server_test.go
```diff
@@ -0,0 +1,344 @@
...
+
+const (
+	postgresHost     = "127.0.0.1"
+	postgresPort     = 5432
+	postgresUser     = "reserve_stats"
+	postgresPassword = "reserve_stats"
+	postgresDatabase = "reserve_stats"
+)
+
+func newTestDB(sugar *zap.SugaredLogger) (*storage.AppNameDB, error) {
```
favadi: please add test case for posting application with same name, it should update the application and returns 200 instead of 500.

app-names/storage/app_name_storage.go
```diff
@@ -0,0 +1,318 @@
...
+const (
+	appNameTable       string = "app_name"
+	addressesTableName string = "addresses"
+)
+
+// ErrAppNotExist exported error for checking
+var ErrAppNotExist = errors.New("app does not exist")
+
+// ErrAddrExisted exported error for checking
+var ErrAddrExisted = errors.New("address already exists")
```
favadi: please group `var()`.

app-names/storage/app_name_storage.go
```diff
@@ -0,0 +1,318 @@
...
+		logger = adb.sugar.With(
+			"func", "appname/storage.CreateOrUpdate",
+			"app name", app.AppName,
+		)
+		appID int64
+	)
+	logger.Debug("Create or update an app")
+	tx, err := adb.db.Beginx()
+	if err != nil {
+		return app, err
```
favadi: please make sure that we called `tx.Commit()` on success path and `tx.RollBack()` on every failure path. This could be done in a defer.

favadi: Also, could you check if we have this problem anywhere else? If yes, create task for each of them.

app-names/storage/app_name_storage.go
```diff
@@ -0,0 +1,318 @@
...
+		}
+		if _, err := tx.Exec(`INSERT INTO addresses (address, app_name_id) VALUES ($1, $2)`, addressStr, appID); err != nil {
+			return updatedApp, err
+		}
+	}
+	if err := tx.Commit(); err != nil {
+		return updatedApp, err
+	}
+	updatedApp, err = adb.GetAppAddresses(appID)
+	return updatedApp, err
```
favadi: just return `adb.GetAppAddresses(appID)`.

app-names/storage/app_name_storage.go
```diff
@@ -0,0 +1,318 @@
...
+	logger.Debug("Get all apps")
+
+	query := `SELECT app_name.id, app_name.name, addresses.address from 
+	app_name JOIN addresses ON app_name.id = addresses.app_name_id WHERE app_name.active IS TRUE `
+
+	if strings.TrimSpace(name) != "" {
+		query += fmt.Sprintf("AND app_name.name='%s'", name)
+	}
+	logger.Debug(query)
+	if err := adb.db.Select(&qresult, query); err != nil {
```
favadi: could this be simplifed by using `db.Scan/ScanStruct` or something?

app-names/storage/app_name_storage.go
```diff
@@ -0,0 +1,318 @@
...
+func (adb *AppNameDB) GetAppAddresses(appID int64) (common.AppObject, error) {
+	var (
+		logger = adb.sugar.With(
+			"func", "appname/storage.GetAppAddresses",
+		)
+		qresult []AppQueryResult
+		result  common.AppObject
+	)
+	logger.Debug("get an app")
+	if err := adb.db.Select(&qresult, fmt.Sprintf(`SELECT app_name.id, app_name.name, addresses.address
```
favadi: same question as above about using `db.Scan`.

app-names/storage/app_name_storage.go
```diff
@@ -0,0 +1,318 @@
...
+	result.AppName = qresult[0].AppName
+	for _, item := range qresult {
+		result.Addresses = append(result.Addresses, ethereum.HexToAddress(item.Address))
+	}
+
+	return result, nil
+}
+
+//DeleteApp set app active is false
+func (adb *AppNameDB) DeleteApp(appID int64) error {
```
favadi: now we decided to go with soft delete, all above GET APIs should fileter the active field, right? Please add test cases for them.

app-names/storage/app_name_storage.go
```diff
@@ -0,0 +1,318 @@
...
+
+	if _, err := tx.Exec("UPDATE app_name SET active=FALSE WHERE id=$1", appID); err != nil {
+		return err
+	}
+	logger.Debug("finish delete app")
+	return tx.Commit()
+}
+
+//DeleteAddress soft delete an address
+func (adb *AppNameDB) DeleteAddress(address string) error {
```
favadi: I don't see this function is used anywhere.

app-names/http/server.go
halink0803: This api for both create and update, if update the request must have id means q.ID != 0, then server return 200 as ok, if create then the request don't have id mean q.ID == 0, the server return 201 as created.

app-names/http/server.go
halink0803: just bad request, no message

app-names/http/server_test.go
halink0803: Name is not defined as an unique key, we use app id to determine which app is updated. If user POST app with same name and same addresses without any id, it will return conflict, not update.

app-names/storage/app_name_storage.go
halink0803: We already have test for delete and get app after delete? What should we test more here? Re-activate?

app-names/http/server_test.go
favadi: We should update application with existing name and returns 200.

app-names/storage/app_name_storage.go
favadi: Also, we should add a query param to GET application-names/ like active=false to list inactive application.

app-names/http/server.go
favadi: please make it returns invalid app id or something meaningful.

app-names/http/server.go
halink0803: Gin is going to support it soon, I add TODO to use it in the future: https://github.com/gin-gonic/gin/pull/1612

app-names/http/server.go
halink0803: Gin is going to support it soon, I add TODO to use it in the future: gin-gonic/gin#1612

favadi:
close in favor of !306.

PR #137, opened 2018-12-07
implement redis for user profile cache

cyberliem:


docker-compose.yml
```diff
@@ -17,4 +17,10 @@ services:
       POSTGRES_USER: reserve_stats
       POSTGRES_PASSWORD: reserve_stats
     ports:
-      - '127.0.0.1:5432:5432'
\ No newline at end of file
+      - '127.0.0.1:5432:5432'
+  redis:
+    image: 'redis:5.0.2-alpine'
+    volumes:
+      - './data/influxdb:/var/lib/redis/'
```
favadi: the data dir is wrong, it should be `/data/redis`.

lib/userprofile/flags.go
```diff
@@ -14,6 +15,10 @@ const (
 	userprofileSigningKeyFlag = "user-profile-signing-key"
 	maxUserCacheFlag          = "max-user-profile-cache"
 	maxUserCacheDefault       = 1000
+	redisEndpointFlag         = "redis-endpoint"
+	redisUserProfileDBFlag    = "redis-user-profile-db"
+	redisUserProfileDBDefault = 0
+	redisPasswordFlag         = "redis-password"
```
favadi: all those flags should make more generic like InfluxDB, PostgresQL as it might useful for other services.

lib/userprofile/flags.go
```diff
@@ -61,10 +84,26 @@ func NewClientFromContext(sugar *zap.SugaredLogger, c *cli.Context) (*Client, er
 }
 
 // NewCachedClientFromContext return new cached client from cli flags
-func NewCachedClientFromContext(client *Client, c *cli.Context) *CachedClient {
+func NewCachedClientFromContext(client *Client, c *cli.Context) (Interface, error) {
```
favadi: this function should accept a redis client instead of `cli.Context`, that makes the input/output clearer.

lib/userprofile/redis_cached_user_profile.go
```diff
@@ -0,0 +1,96 @@
...
+
+func (cc *RedisCachedClient) cacheRedis(addr ethereum.Address, p UserProfile) error {
+	//empty result
+	cc.sugar.Debugf("get user profile from API", "user profile", p)
+	if (p == UserProfile{}) {
+		return cc.redisClient.Set(addr.Hex(), "", 24*time.Hour).Err()
+	}
+
+	resultStr := fmt.Sprintf("%s:%s", p.UserName, strconv.FormatInt(p.ProfileID, 10))
+	return cc.redisClient.Set(addr.Hex(), resultStr, 0).Err()
```
favadi: this does not look right. We should use `hash` data type of redis, see https://redis.io/commands/hmset.

lib/userprofile/redis_cached_user_profile.go
```diff
@@ -0,0 +1,96 @@
...
+		logger.Debugw("cache to redis failed", "redis error", err)
+	}
+	return p, err
+}
+
+func (cc *RedisCachedClient) cacheRedis(addr ethereum.Address, p UserProfile) error {
+	//empty result
+	cc.sugar.Debugf("get user profile from API", "user profile", p)
+	if (p == UserProfile{}) {
+		return cc.redisClient.Set(addr.Hex(), "", 24*time.Hour).Err()
```
favadi: `24h*time.Hour` should be a const somewhere.

lib/userprofile/redis_cached_user_profile.go
```diff
@@ -0,0 +1,96 @@
...
+	err = cc.cacheRedis(addr, p)
+	if err != nil {
+		logger.Debugw("cache to redis failed", "redis error", err)
+	}
+	return p, err
+}
+
+func (cc *RedisCachedClient) cacheRedis(addr ethereum.Address, p UserProfile) error {
+	//empty result
+	cc.sugar.Debugf("get user profile from API", "user profile", p)
```
favadi: please add log context (function name, input params).

lib/userprofile/flags.go
favadi: they must be placed somewhere that other service can import the flag set to use later. For the database name, please see the NewPostgreSQLFlags function.

favadi: @liem this is still unresolved.

lib/userprofile/flags.go
```diff
@@ -60,11 +83,24 @@ func NewClientFromContext(sugar *zap.SugaredLogger, c *cli.Context) (*Client, er
...
-// NewCachedClientFromContext return new cached client from cli flags
-func NewCachedClientFromContext(client *Client, c *cli.Context) *CachedClient {
-	if client == nil {
-		return nil
+//NewRedisClientFromContext creates redis client from flag agruments
+func NewRedisClientFromContext(c *cli.Context) (*redis.Client, error) {
+	redisURL := c.String(redisEndpointFlag)
+	if redisURL == "" {
+		return nil, nil
 	}
```
favadi: we should returns an error when redisURL is empty. The idea is the constructor should returns a fully-ready-to-use instance. Could you check if we did the same anywhere? If yes, I will open a task for each of them.

lib/userprofile/flags.go
```diff
@@ -60,11 +83,24 @@ func NewClientFromContext(sugar *zap.SugaredLogger, c *cli.Context) (*Client, er
...
+		Addr:     redisURL,
+		Password: c.String(redisPasswordFlag),
+		DB:       c.Int(redisUserProfileDBFlag),
+	})
+	_, err := redisClient.Ping().Result()
+	return redisClient, err
+}
+
+//NewInmemCachedFromContext create the inmem cache client from flag agruments
+func NewInmemCachedFromContext(client *Client, c *cli.Context) Interface {
```
favadi: Inmem --> InMem.

lib/userprofile/redis_cached_user_profile.go
```diff
@@ -0,0 +1,102 @@
...
+
+// NewRedisCachedClient creates a new User Profile cached client instance.
+func NewRedisCachedClient(client *Client, redisClient *redis.Client) *RedisCachedClient {
+	return &RedisCachedClient{
+		Client:      client,
+		redisClient: redisClient,
+	}
+}
+
+// LookUpCache will lookup the Userprofile from cache
```
favadi: Userprofile --> UserProfile.

lib/userprofile/redis_cached_user_profile.go
```diff
@@ -0,0 +1,102 @@
...
+// Redis cache will return error if the addr is not exist.
+func (cc *RedisCachedClient) LookUpCache(addr ethereum.Address) (UserProfile, error) {
+	ps, err := cc.redisClient.HMGet(addr.Hex(), nameField, idField).Result()
+	if err != nil {
+		return UserProfile{}, err
+	}
+	if len(ps) < 2 {
```
favadi: we should assert != 2.

lib/userprofile/redis_cached_user_profile.go
```diff
@@ -0,0 +1,102 @@
...
+	p, err := cc.LookUpCache(addr)
+	if err == nil {
+		logger.Debugw("cache hit")
+		return p, nil
+	}
+
+	logger.Debugw("cache missed", "redis error", err)
+
+	p, err = cc.Client.LookUpUserProfile(addr)
+	if err != nil {
```
favadi: L75-76 should be grouped in one line.

lib/userprofile/redis_cached_user_profile.go
```diff
@@ -0,0 +1,102 @@
...
+		return p, err
+	}
+
+	//cache the result into redis and return result
+	err = cc.cacheRedis(addr, p)
+	if err != nil {
```
favadi: L81-82 should be grouped in one line.

lib/userprofile/redis_cached_user_profile.go
```diff
@@ -0,0 +1,102 @@
...
+		logger.Debugw("cache to redis failed", "redis error", err)
+	}
+	return p, err
+}
+
+func (cc *RedisCachedClient) cacheRedis(addr ethereum.Address, p UserProfile) error {
+	//empty result
```
favadi: this comment is weird.

lib/userprofile/redis_cached_user_profile.go
```diff
@@ -0,0 +1,102 @@
...
+	cc.sugar.Debugf("cached user profile from API into redis", "address", addr.Hex(), "user profile", p)
```
favadi: should be `Debugw` as we don't use any formatting here.

lib/userprofile/redis_cached_user_profile.go
```diff
@@ -0,0 +1,102 @@
...
+	resultMap := map[string]interface{}{
+		nameField: p.UserName,
+		idField:   p.ProfileID,
+	}
+	if _, err := cc.redisClient.HMSet(addr.Hex(), resultMap).Result(); err != nil {
+		return err
+	}
+	if (p == UserProfile{}) {
```
favadi: we should turns the Lookup function to returns pointer so we don't have to allocate new instance.

tradelogs/cmd/trade-logs-api/main.go
```diff
@@ -103,3 +102,22 @@ func main() {
 		log.Fatal(err)
 	}
 }
+
+func createUserprofileClient(sugar *zap.SugaredLogger, c *cli.Context) (userprofile.Interface, error) {
```
favadi: the constructor should be `newSomething`.

tradelogs/cmd/trade-logs-api/main.go
```diff
@@ -103,3 +102,22 @@ func main() {
...
+	if err != nil {
+		return nil, err
+	}
+
+	redisClient, err := userprofile.NewRedisClientFromContext(c)
+	if err != nil {
+		return nil, err
+	}
+
+	if redisClient == nil {
```
favadi: this check should be replaced with `c.String(redisURL) == ""`.

lib/userprofile/flags.go
cyberliem: In this PR i define if redisURL is empty, the service will use normal caching service.
As in other service, for example, PostGres, we didn't check empty URL
https://github.com/KyberNetwork/reserve-stats/blob/524f4f6948e93fb7990e1cd37706012c3bcd1ce9/lib/app/postgresql.go#L64

lib/userprofile/redis_cached_user_profile.go
cyberliem: this is dangerous since we're handling empty result as well, the endpoint did not specify an error or anything, just return an empty json string if there is no matched record.

tradelogs/cmd/trade-logs-api/main.go
cyberliem: redisURLFlag is on a different packages.

lib/userprofile/flags.go
favadi: we will fix Postgres too, I'll open a task. Consumer should be able to use new instance returned by constructor and expect it to work, otherwise we will need to check if it is nil everywhere.

lib/userprofile/redis_cached_user_profile.go
favadi: Not much of a problem, we can keep this.

tradelogs/cmd/trade-logs-api/main.go
cyberliem: if redis constructor returns an error that means the user want to use redis cache but the init can not be finished. It's different with user does not want to use redis cache hence the redis client == nil

tradelogs/cmd/trade-logs-api/main.go
```diff
@@ -103,3 +102,22 @@ func main() {
 		log.Fatal(err)
 	}
 }
+
+func newUserprofileClient(sugar *zap.SugaredLogger, c *cli.Context) (userprofile.Interface, error) {
```
favadi: `newUserprofileClient` --> `newUserProfileClient`.

lib/app/redis.go
```diff
@@ -0,0 +1,52 @@
...
+		},
+	}
+}
+
+//NewRedisClientFromContext creates redis client from flag agruments
+func NewRedisClientFromContext(c *cli.Context) (*redis.Client, error) {
+	redisURL := c.String(redisEndpointFlag)
+	if redisURL == "" {
+		return nil, nil
+	}
```
favadi: The constructor still returns nil value. We should returns an error like "no redis endpoint in context" or something.

cyberliem: As discussed here, https://github.com/KyberNetwork/reserve-stats/pull/137#discussion_r242064827, I think the error return is illogical. 

In case there must be an error added, I will add another flag to let user explicitly defined if they want redis as cache for user.

PR #139, opened 2018-12-12
1610 enumerated crawled data

cyberliem:


tradelogs/storage/influx.go
```diff
@@ -104,27 +110,92 @@ func (is InfluxStorage) LastBlock() (int64, error) {
...
+			logschema.SrcAmount,
+			logschema.DstAmount,
+			logschema.IP,
+			logschema.Country,
+			logschema.IntegrationApp,
+		}
+		tradeLogQuery string
+	)
+	for _, field := range tradeLogQueryFields {
+		tradeLogQuery += field.String() + ", "
```
favadi: could we use `strings.Join`?

tradelogs/storage/influx.go
```diff
@@ -103,27 +109,92 @@ func (is InfluxStorage) LastBlock() (int64, error) {
...
+			burnschema.Time,
+			burnschema.ReserveAddr,
+			burnschema.Amount,
+			burnschema.LogIndex,
+		}
+		burnFeeQuery string
+	)
+	for i, field := range burnFeeFields {
+		if i != 0 {
+			burnFeeQuery += ", "
```
favadi: could we use strings.Join?

tradelogs/storage/influx.go
```diff
@@ -103,27 +109,92 @@ func (is InfluxStorage) LastBlock() (int64, error) {
...
+			walletschema.ReserveAddr,
+			walletschema.WalletAddr,
+			walletschema.Amount,
+			walletschema.LogIndex,
+		}
+		walletQuery string
+	)
+	for i, field := range walletFeeFields {
+		if i != 0 {
+			walletQuery += ", "
```
favadi: could we use strings.Join?

tradelogs/storage/influx.go
```diff
@@ -103,27 +109,92 @@ func (is InfluxStorage) LastBlock() (int64, error) {
...
+			logschema.IP,
+			logschema.Country,
+			logschema.IntegrationApp,
+		}
+		tradeLogQuery string
+	)
+	for _, field := range tradeLogQueryFields {
+		tradeLogQuery += field.String() + ", "
+	}
+	fiatAmount := fmt.Sprintf("(%s + %s) AS %s", logschema.EthAmount.String(), logschema.EthUSDRate.String(), logschema.FiatAmount.String())
```
favadi: the operator should be * instead of +, right?

tradelogs/storage/influx.go
cyberliem: we can but the time required to correct this outweigh the advantages.

tradelogs/storage/influx.go
cyberliem: Fixed. I did the following test:

- 1. Drop Database.
- 2. Crawled from block 6699900 to 6700000
- 3. Query Tradelogs JSON from API. 
- 4. Switch back to Develop branch, re run it.
- 5. Compare both JSON.

The result came identical, so this bug is not going to happen again.

tradelogs/storage/influx.go
cyberliem: we can but the time required to correct this outweigh the advantages.

PR #141, opened 2018-12-17
1937 Update lastrate return multiple pair

halink0803:


reserverates/storage/influx/rates_influxdb.go
```diff
@@ -110,87 +110,94 @@ func (rs *RateStorage) LastBlock() (int64, error) {
...
+			schema.Pair.String(),
+			pair)
+
+		logger.Debugw("query statement", "query", stmt)
+
+		q := influxClient.NewQuery(stmt,
+			rs.dbName,
+			timePrecision)
+
+		res, err := rs.client.Query(q)
```
favadi: I don't see how it is an improvement, we still query multiple times. Please investigate if it is possible to query once.

halink0803: Yes, it is possible to query once. I did update the function. Please check.

PR #144, opened 2018-12-20
1927 Fix not commit before return

halink0803:


users/migration/migrate_test.go
```diff
@@ -70,6 +71,17 @@ func newTestMigrateDB() (*DBMigration, error) {
...
+		if err != nil {
+			rollBackErr := tx.Rollback()
+			if rollBackErr != err {
+				logger.Debug(fmt.Sprintf("rollback error: %s", rollBackErr))
+			}
+			return
+		}
+		err = tx.Commit()
+	}()
+
```
cyberliem: line[89](https://github.com/KyberNetwork/reserve-stats/pull/144/files#diff-d36a660d7769c228f51024d770c0cea9R89) double commit again.

cyberliem: [deletedb](https://github.com/KyberNetwork/reserve-stats/pull/144/files#diff-d36a660d7769c228f51024d770c0cea9R26) should this be done without tx?

halink0803: @cyberliem how should it be done?

PR #146, opened 2018-12-20
2016 token symbol from blockchain

ductm54:


lib/contracts/erc20_type2.go
```diff
@@ -0,0 +1,676 @@
+// Code generated - DO NOT EDIT.
```
favadi: we need to add this in abi_gen.sh file, right?

favadi: Also, please add a comment to tell what is type 1 and type 2.

lib/core/test/symbols_test.go
```diff
@@ -0,0 +1,140 @@
...
+				resourcesCh <- struct{}{}
+				defer func() { <-resourcesCh }()
+
+				symbols, err := getSymbol(address, client)
+				if err != nil {
+					symbols, err = getSymbol2(address, client)
+					if err != nil {
+						return err
+					}
+				}
```
favadi: this part of codes could be clearer with something like that.

```go
var getSymbolFns = []func(address, client) symbols, err // we could define a type for this

var symbol
for _, fn := range getSymbolFns {
    if symbol, err = fn(address, client); err != nil {
        // log the err
    }
    break
}

if err != nil {}
```

favadi: Also, is it possible to catch only the error when it is failed to unpack the symbol to string?

PR #150, opened 2018-12-24
1869 fix logic fault in user migration

cyberliem:


users/migration/migrate.go
```diff
@@ -133,10 +135,10 @@ func (dbm *DBMigration) insertIntoUserTable(email string) (int, error) {
 		}
 	} else if err != nil {
 		return 0, err
+	} else if err == nil {
```
favadi: do we really need explicit `err = nil`? because the right before case is `err !=nil` already. Also, do the log message really need to move into the if-else clause?

users/migration/migrate.go
```diff
@@ -146,6 +148,27 @@ func (dbm *DBMigration) insertIntoUserTable(email string) (int, error) {
...
+	)
+
+	err = ptx.Get(&userIDFromDB, fmt.Sprintf(`SELECT user_id FROM "%s" WHERE address = $1;`, addressesTableName), address)
+	if err == nil {
+		if userID != userIDFromDB {
+			logger.Debugw("address already belong to a different user", "userID input", userID, "userID from DB", userIDFromDB)
+			if err = ptx.Commit(); err != nil {
+				return err
+			}
+			return errors.New("address is already existed with a different userID")
```
favadi: why do we need this check? address field is not unique already?

cyberliem: Since it's unique, Pql raised error when we try to update an address that is already exist.

I can catch this error and ignore it if we got duplicate address update. However consider the verbal discussion between us a day ago, an address must belong to the same User. If we try to update it with a different user I think we should raise an error.

users/migration/migrate.go
cyberliem: if err==sql.ErrNoRows() in line 130m, this condition is not guaranteed to return. The code might exit this if clause and printed out "user already exists, skipping" although the user is not exist. Rerun the develop build to see this wrong Log printing.

users/migration/migrate.go
favadi: got it, so what you are trying to do are:

- if the address already exist with same user_id, ingoring it
- if the address already exist but with different user_id, return an error

That sounds reasonable, but we should extend it further by not allow any conflict, just update the INSERT statement to only execute with condition that `NOT EXIST SELECT * FROM addresses WHERE address=xx,id=yyy`.

users/migration/migrate.go
```diff
@@ -146,6 +148,26 @@ func (dbm *DBMigration) insertIntoUserTable(email string) (int, error) {
...
+	var (
+		logger = dbm.sugar.With(
+			"func", "users/migration/DBMigration.insertAddress",
+			"address", address,
+		)
+		userIDFromDB int
+	)
+
+	err = ptx.Get(&userIDFromDB, fmt.Sprintf(`SELECT user_id FROM "%s" WHERE address = $1;`, addressesTableName), address)
+	if err == nil {
```
favadi: the err != nil should be handle first and the code blocks inside err = nil should be move to main flow.

users/migration/migrate.go
favadi: > As we do not enforce it in logic of users db update, I don't know if we should enforce it in migrate. For now I'll just let it overwrite the old user_ID with new user_ID

The different here is there should not be any already exists address at all in database migration, as the POST handler will override them already.

> I don't know if we should enforce it in migrate

We should if it is easy to do, otherwise, nvm.

users/migration/migrate.go
cyberliem: Fixed, else part is not without the clause err==nil

users/migration/migrate.go
cyberliem: To change this, we can alter the schema user_id to unique and raise error on conflict. However I don't think this is necessary.

users/migration/migrate.go
```diff
@@ -71,6 +71,51 @@ CREATE TABLE IF NOT EXISTS "%[2]s" (
...
+
+	tx, err := dbm.postgresdb.Beginx()
+	if err != nil {
+		return err
+	}
+
+	defer pgsql.CommitOrRollback(tx, logger, &err)
+	timepoint := timeutil.TimestampMsToTime(timestamp)
+	stmt = fmt.Sprintf(`WITH u AS (
+  INSERT INTO "%[1]s" (email, last_updated)
```
favadi: the last_updated should be the timestamp when a user record is updated in database, useful for debugging purpose. Please change it to just now().

cyberliem: fixed it in commit https://github.com/KyberNetwork/reserve-stats/pull/150/commits/733a5bc2747b3ba8898e066f9af1dbd32e67058b

PR #152, opened 2018-12-25
re check KYC status for trade_logs and re-execute KYC related metric

cyberliem:


tradelogs/cmd/kyc-reaggregate/main.go
```diff
@@ -0,0 +1,112 @@
...
+	"github.com/urfave/cli"
+	"go.uber.org/zap"
+)
+
+const (
+	defaultDB    = "reserve_stats"
+	fromTimeFlag = "fromTime"
+	toTimeFlag   = "toTime"
+	//This is 01 Feb 2018 UTC
+	defaultFromTime = 1517443200000
```
favadi: could this be replaced with `SELECT FIRST(something)` in database?

tradelogs/cmd/kyc-reaggregate/main.go
```diff
@@ -0,0 +1,112 @@
...
+
+func main() {
+	app := libapp.NewApp()
+	app.Name = "Trade Logs KYC re-aggregate"
+	app.Version = "0.0.1"
+	app.Action = run
+	app.Flags = append(app.Flags,
+		cli.Uint64Flag{
+			Name:   fromTimeFlag,
+			Usage:  "re-KYC from time. Default is 00:00:00",
```
favadi: I think the default time is in usage is misleading.

favadi: Also could the usage state something about format if from/to time (in millis/micro/nano)?

tradelogs/cmd/kyc-reaggregate/main.go
```diff
@@ -0,0 +1,112 @@
...
+			EnvVar: "FROM_TIME",
+			Value:  defaultFromTime,
+		},
+		cli.Uint64Flag{
+			Name:   toTimeFlag,
+			Usage:  "re-KYC to time",
+			EnvVar: "TO_TIME",
+			Value:  0,
```
favadi: could the to time default to `SELECT LAST(something)` from database?

tradelogs/cmd/kyc-reaggregate/main.go
```diff
@@ -0,0 +1,112 @@
+package main
```
favadi: kyc_reaggregate --> Please change the name of the cmd to `trade-logs-kyc-updater`.

tradelogs/workers/rekyc_workers.go
```diff
@@ -0,0 +1,72 @@
...
+// ReKYCJob describe a
+type ReKYCJob struct {
+	c     *cli.Context
+	order int
+	from  time.Time
+	to    time.Time
+}
+
+// NewReKYCJob returns a ReKYCJob
+func NewReKYCJob(c *cli.Context, order int, from, to time.Time) *ReKYCJob {
```
favadi: it seems that order is not in used anywhere.

tradelogs/workers/rekyc_workers.go
```diff
@@ -0,0 +1,72 @@
...
+		to:    to,
+	}
+}
+
+// Execute get the tradelogs and re-kyc it
+func (rk *ReKYCJob) Execute(sugar *zap.SugaredLogger) error {
+	return rk.getAndReKYC(sugar)
+}
+
+func (rk *ReKYCJob) getAndReKYC(sugar *zap.SugaredLogger) error {
```
favadi: the function could be called `update`.

tradelogs/cmd/kyc-reaggregate/main.go
cyberliem: If we want to make sure, we can make this 0, it will guarantee to look from the beginning. (1970).

PR #156, opened 2018-12-28
1479 Check worker number

halink0803:


tradelogs/cmd/trade-logs-crawler/main.go
```diff
@@ -278,6 +279,10 @@ func run(c *cli.Context) error {
 			sugar.Infow("fetching trade logs up to latest known block number", "to_block", toBlock.String())
 		}
 
+		if fromBlock.Int64() >= toBlock.Int64() {
```
favadi: could we use native big.Int compare operator directly.

PR #157, opened 2018-12-28
2069 Using address instead of token symbol

halink0803:


tradelogs/http/asset_volume_handler.go
```diff
@@ -3,15 +3,15 @@ package http
...
 	"github.com/KyberNetwork/reserve-stats/lib/httputil"
 	_ "github.com/KyberNetwork/reserve-stats/lib/httputil/validators" // import custom validator functions
+	"github.com/ethereum/go-ethereum/common"
 	"github.com/gin-gonic/gin"
 )
 
 type assetVolumeQuery struct {
 	httputil.TimeRangeQueryFreq
-	Asset string `form:"asset" binding:"required"`
+	Asset string `form:"asset" binding:"required,isAddress"`
```
favadi: we need to update the API docs?

tradelogs/http/asset_voume_http_test.go
```diff
@@ -71,7 +71,7 @@ func TestAssetVolumeHttp(t *testing.T) {
 		invalidFrom = "xxxx"
 		validTo     = 1539302400000
 		// mock core only return ETH, KNC is not in the list of mock core's clients
-		validAsset               = "ETH"
+		validAsset               = "0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee"
```
favadi: we have const with this address somewhere already, I think.

tradelogs/http/asset_volume_handler.go
halink0803: Updated the api docs

PR #159, opened 2019-01-02
2107 contract v3 kyber trade event

cyberliem:


tradelogs/crawled-log-utils.go
```diff
@@ -0,0 +1,82 @@
+package tradelogs
```
favadi: we use _ instead of - for file name everywhere, please rename crawled-log-utils.go --> crawled_log_utils.go. Also, is it a good name?

tradelogs/crawler.go
```diff
@@ -30,6 +30,14 @@ const (
...
 	// etherReceivalEvent is the topic of event EtherReceival(address indexed sender, uint amount).
 	etherReceivalEvent = "0x75f33ed68675112c77094e7c5b073890598be1d23e27cd7f6907b4a7d98ac619"
+	// kyberTradeEvent is the topic of event
+	// KyberTrade (address trader, address src, address dest, uint256 srcAmount, uint256 dstAmount, address destAddress, uint256 ethWeiValue, address reserve1, address reserve2, bytes hint)
+	// ETHwei = ETHAMOUNT
+	// rsv1 ==0 if eth -> token
+	// rsv2 ==0 if token -> eth
+	kyberTradeEvent = "0xd30ca399cb43507ecec6a629a35cf45eb98cda550c27696dcb0d8c4a3873ce6c"
+	// startingBlockV3 is the block where V3 contract is used.
+	startingBlockV3 = 6997111
```
favadi: the starting block for V3 will be different between deployments, we need to do the samething for contract address.

tradelogs/crawled-log-utils.go
```diff
@@ -0,0 +1,82 @@
...
+	"github.com/ethereum/go-ethereum/core/types"
+)
+
+func (crawler *Crawler) handleEventLogWithBlockNumber(log types.Log, tradeLog *common.TradeLog) (bool, error) {
+	//this should not happen, but still...
+	if len(log.Topics) == 0 {
+		return false, errors.New("log item has no topic")
+	}
+	//if blockNumber < startingBlock/
+	if startingBlockV3 == 0 || log.BlockNumber < startingBlockV3 {
```
favadi: there is no improvement if we still listen for all events and process all topics. We should only listen to KyberTrade, burnFee if block >= starting V3. To make it clear, we should have 2 functions (the function name is just for demonstrate purpose, invent one yourself):

- fetchLogsV1
- fetchLogsV2
- fetchLogsV3 

while V1, V2 point to a same function that contains current logs fetching logic. The V3 fetchLogsV3 one should listen to its **required** address, topics (no internal_network addresss is needed for V3).

tradelogs/crawled-log-utils.go
cyberliem: Originally I intended to move all "logDataTo..." functions into this file but then in the end I think I should not be changing anything out-of-scope

tradelogs/versioned_log_handler.go
```diff
@@ -0,0 +1,148 @@
...
+			return true, nil
+			// prepare to fulfill new TradeLog from next event logs
+		default:
+			crawler.sugar.Info("Unknown log topic.")
+			return false, nil
+		}
+	}
+
+	// We're not picking up reserver Addressese from KyberTrade event, they will be concerned at burnfee
+	if log.Topics[0].Hex() == kyberTradeEvent {
```
favadi: not sure about the comment, but we should only listen for KyberTradeEvent if block > V3

tradelogs/versioned_log_handler.go
```diff
@@ -0,0 +1,148 @@
...
+	topics := [][]ethereum.Hash{
+		{
+			ethereum.HexToHash(tradeEvent),
+			ethereum.HexToHash(burnFeeEvent),
+			ethereum.HexToHash(feeToWalletEvent),
+			ethereum.HexToHash(etherReceivalEvent),
+		},
+	}
+
+	typeLogs, err := crawler.fetchTypeLogsWithTopics(fromBlock, toBlock, timeout, topics)
```
favadi: what is `typeLogs`? :)

tradelogs/versioned_log_handler.go
```diff
@@ -0,0 +1,148 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	result, err = crawler.assembleTradeLogs(typeLogs)
```
favadi: we should have `fetchTradeLogV2` and `assembleTradeLogsV3` just like `fetchTradeLogV2` and `fetchTradeLogV3`. Also, as we are already refactoring, we should make it clearer by:

- fetchTradeLogV2 input
  - etherReceivalEvent (optional)
  - tradeEvent [required]
  - feeToWalletEvent (optional)
  - burnFeeEvent (optional)
- fetchTradeLogV2 inputs
  - kyberTrade (optional)
  - feeToWalletEvent (optional)
  - burnFeeEvent (optional)

the shared logic between 2 function should be grouped in a function.

Also, we

favadi:
I feel like the logic for v2, v3 should be split to separate files like crawler_v2.go, crawler_v3.go.

PR #171, opened 2019-01-07
2147 add knc fetcher and cmd

cyberliem:


lib/timeutil/flags.go
```diff
@@ -0,0 +1,52 @@
...
+	"errors"
+	"time"
+
+	"github.com/urfave/cli"
+)
+
+const (
+	toTimeFlag        = "to"
+	fromTimeFlag      = "from"
+	timePrecisionFlag = "time-precision"
```
favadi: unused.

lib/timeutil/flags.go
```diff
@@ -0,0 +1,52 @@
...
+)
+
+// NewTimeRangeCliFlags returns cli flags to configure a fromTime-toTime.
+func NewTimeRangeCliFlags() []cli.Flag {
+	return []cli.Flag{
+		cli.Uint64Flag{
+			Name:   fromTimeFlag,
+			Usage:  "from time. Default is  00:00:00 UTC on 1 January 1970 in millisecond",
+			EnvVar: "FROM_TIME",
+			Value:  0,
```
favadi: as we are fetching daily data, there is no point accepting second precision time range input (we will need to round to day any way). Please change the flag to accept something like `--from '2018-01-2' --to '2019-01-02'`. It can be achieve by parsing the string flat with `time.Parse` with time layout `2006-01-02`. The usage of the string flags should describe the time format.

lib/tokenrate/token_rate_fetcher.go
```diff
@@ -0,0 +1,137 @@
...
+		}
+		res = response.Results
+	} else {
+		return res, err
+	}
+	return res, nil
+}
+
+//FetchRatesInRanges fetch and saves times into influxDB
+func (rf *RateFetcher) FetchRatesInRanges(from, to time.Time, tokenID, currency string) error {
```
favadi: all of those are very specific to knc-usd-rate service, I don't see how it can be reuse for other services. We should:

- rename `knc-usd-rate` to a more generic name `token-rate-crawler`
- move this codes inside the new package
- for now just need to hard code `knc / usd` pair in main.go, we might support other pairs in the future.

lib/tokenrate/token_rate_fetcher.go
```diff
@@ -0,0 +1,137 @@
...
+			rf.sugar.Errorw("Failed to get influx point from rate", "error", err)
+			return err
+		}
+		bp.AddPoint(pt)
+	}
+	return rf.influxClient.Write(bp)
+}
+
+//FetchRates return the rate for pair token-currency at timestamp timeStamp
+func (rf *RateFetcher) FetchRates(token, currency string, timeStamp time.Time) (TokenRate, error) {
```
favadi: this function could be unexported.

lib/tokenrate/token_rate_fetcher.go
```diff
@@ -0,0 +1,137 @@
...
+	var (
+		USDRate float64
+		rate    TokenRate
+		err     error
+	)
+	if USDRate, err = rf.rateProvider.Rate(token, currency, timeStamp); err != nil {
+		rf.sugar.Errorw(fmt.Sprintf("failed to get %s/%s rate", token, currency), "timestamp", timeStamp.String())
+		return rate, err
+	}
+	if USDRate != 0 {
```
favadi: what if `USDRate == 0`? we will save a all-nil token-rate to database? I think we should just honor the rate returned by provider.

lib/tokenrate/token_symbol.go
```diff
@@ -0,0 +1,15 @@
+package tokenrate
+
+//tokenSymbol is the mapping of tokenID to token Symbol
+var tokenSymbols = map[string]string{
+	"kyber-network": "knc",
```
favadi: the id is very specific to coingecko provider. Also, this should be moved to the token-rate-fetcher service.

favadi:
We need add the logic to cmd that:

- start fetching from last stored date in database if database is not empty
- if `--to` is not provided, run in daemon mode that sleep for few hours and fetch all rates from last stored day to now

PR #172, opened 2019-01-08
2165 - Update nonce_validator.go

ductm54:


gateway/http/nonce_validator.go
```diff
@@ -2,6 +2,7 @@ package http
 
 import (
 	"errors"
+	"fmt"
 	"github.com/KyberNetwork/reserve-stats/lib/timeutil"
```
halink0803: Please separate internal packages and external packages

PR #185, opened 2019-01-14
2179  Add wait-for-confirmations flag

halink0803:


tradelogs/cmd/trade-logs-crawler/main.go
```diff
@@ -269,7 +279,7 @@ func run(c *cli.Context) error {
 			if fErr != nil {
 				return fErr
 			}
-			toBlock = currentHeader.Number
+			toBlock = currentHeader.Number.Sub(currentHeader.Number, big.NewInt(blockConfirmation))
```
favadi: I can think of a scenario like that:

- from-block: 100
- to-block: 102

crawler will try to crawl from 100 --> 102 - 7, from-block < to-block --> exit. Could you double check?

halink0803: Yes, I already check it. The logic already implement here:
https://github.com/KyberNetwork/reserve-stats/blob/a60ba3754e42d41af4372e1fad11b890e3ca62c7/tradelogs/cmd/trade-logs-crawler/main.go#L286

favadi: but that means the program will returns an error and exit?

halink0803: @favadi I change the logic to sleep(delayTime) when latest block - block confirmations is smaller fromBlock.

PR #188, opened 2019-01-15
2221 Add reserves to weth eth logs

halink0803:


tradelogs/crawler_v2.go
```diff
@@ -34,6 +40,30 @@ func (crawler *Crawler) fetchTradeLogV2(fromBlock, toBlock *big.Int, timeout tim
...
+		return reserveAddr, err
+	}
+	for _, log := range receipt.Logs {
+		for _, topic := range log.Topics {
+			if topic == ethereum.HexToHash(tradeExecuteEvent) {
+				reserveAddr = log.Address
+				break
+			}
+		}
+		if reserveAddr.Hex() != "0x0000000000000000000000000000000000000000" {
```
favadi: could this be replaced with len(reserveAddr) == 0, or at least we should have something like isAddressZero somewhere.

tradelogs/common/types.go
```diff
@@ -36,9 +42,11 @@ type TradeLog struct {
 	UserAddress ethereum.Address `json:"user_addr"`
 	SrcAddress  ethereum.Address `json:"src_addr"`
 	DestAddress ethereum.Address `json:"dst_addr"`
-	SrcAmount   *big.Int         `json:"src_amount"`
-	DestAmount  *big.Int         `json:"dst_amount"`
-	FiatAmount  float64          `json:"fiat_amount"`
+	//ReserveAddresses is reserve address for case when burn fee is nil
+	ReserveAddresses ReserveAddress `json:"reserve_addr"`
```
favadi: I think we should just place src_reserve_addr and dst_reserve_addr directly to the TradeLog struct.

tradelogs/crawler_v2.go
```diff
@@ -71,6 +101,15 @@ func (crawler *Crawler) assembleTradeLogsV2(eventLogs []types.Log) ([]common.Tra
 			if tradeLog.Timestamp, err = crawler.txTime.Resolve(log.BlockNumber); err != nil {
 				return nil, err
 			}
+			// when the tradelog does not contain burnfee and etherReceival event
+			// get tx receipt to get reserve address
+			if len(tradeLog.BurnFees) == 0 && tradeLog.EtherReceivalSender.Hex() == "0x0000000000000000000000000000000000000000" {
```
favadi: could this be replaced with len(reserveAddr) == 0, or at least we should have something like isAddressZero somewhere.

tradelogs/storage/influx.go
```diff
@@ -342,6 +344,14 @@ func (is *InfluxStorage) tradeLogToPoint(log common.TradeLog) ([]*client.Point,
 		} else {
 			logger.Warnw("unexpected burn fees", "got", log.BurnFees, "want", "1 burn fees (dst)")
 		}
+	} else {
+		if log.EtherReceivalSender.Hex() != "0x0000000000000000000000000000000000000000" {
+			tags[logschema.SrcReserveAddr.String()] = log.EtherReceivalSender.Hex()
+		} else if log.ReserveAddresses.SrcReserveAddress.Hex() != "0x0000000000000000000000000000000000000000" {
```
favadi: we really need a common isZero function.

tradelogs/crawler_v2.go
```diff
@@ -34,6 +44,30 @@ func (crawler *Crawler) fetchTradeLogV2(fromBlock, toBlock *big.Int, timeout tim
...
+	defer cancel()
+	receipt, err := crawler.ethClient.TransactionReceipt(ctx, txHash)
+	if err != nil {
+		return reserveAddr, err
+	}
+	for _, log := range receipt.Logs {
+		for _, topic := range log.Topics {
+			if topic == ethereum.HexToHash(tradeExecuteEvent) {
+				reserveAddr = log.Address
+				break
```
cyberliem: What if there is more than one trade in that TX? What do you think about using log index in this lookup to ensure the tradeExecuteEvent belong exactly to the trade happen in ExecuteTrade followed?

tradelogs/storage/influx.go
```diff
@@ -342,6 +344,14 @@ func (is *InfluxStorage) tradeLogToPoint(log common.TradeLog) ([]*client.Point,
 		} else {
 			logger.Warnw("unexpected burn fees", "got", log.BurnFees, "want", "1 burn fees (dst)")
 		}
+	} else {
+		if log.EtherReceivalSender.Hex() != "0x0000000000000000000000000000000000000000" {
```
cyberliem: How about exporting the function IsZeroAddress and use it for these conditions?  It could improve the readability

tradelogs/crawler_v2.go
```diff
@@ -71,6 +105,15 @@ func (crawler *Crawler) assembleTradeLogsV2(eventLogs []types.Log) ([]common.Tra
 			if tradeLog.Timestamp, err = crawler.txTime.Resolve(log.BlockNumber); err != nil {
 				return nil, err
 			}
+			// when the tradelog does not contain burnfee and etherReceival event
+			// get tx receipt to get reserve address
+			if len(tradeLog.BurnFees) == 0 && isZeroAddress(tradeLog.EtherReceivalSender) {
```
cyberliem: Does it make sense to assign tradeLog.SrcReserveAddress = tradeLog.EtherReceivalSender in case tradeLog.BurnFees is empty and EtherReceivalSender is not zero?

Also, does it make sense to assign tradeLog.SrcReseveAddress and DstReserveAddress from burnFee events in fillBurnFees?

tradelogs/storage/influx.go
cyberliem: Also Will it be better if we ensure the logic of DstAndSrc Reserve Address from crawler, and save ourselves from these unnecessary nested condition?

In essence we can just assign tag[srcRsvAddr]=log.SrcRsvAddr.Hex() and tag[dstRsvAddr]=log.DstRsvAddr.Hex(). The implementation of translating burnFees arrays to these params are obsolete since we already know there can only be at most 2 burn fee events in each trade

tradelogs/crawler_v2.go
halink0803: @cyberliem If it does not what could be reserve address then?

tradelogs/storage/influx.go
halink0803: @cyberliem since crawler we crawl by topics then we are not sure if the log had  burnFee or not.

The next comment I do not really understand but if you mean we move burnFee into tradeLogs then we can do it in another task.

lib/contracts/addresses.go
```diff
@@ -5,6 +5,11 @@ import (
 	"github.com/ethereum/go-ethereum/common"
 )
 
+//IsZeroAddress return if address is zero
+func IsZeroAddress(address common.Address) bool {
```
favadi: this should be moved to blockchain package.

tradelogs/crawler_v2.go
favadi: @cyberliem has a valid point here.

tradelogs/storage/influx.go
```diff
@@ -342,6 +345,14 @@ func (is *InfluxStorage) tradeLogToPoint(log common.TradeLog) ([]*client.Point,
 		} else {
 			logger.Warnw("unexpected burn fees", "got", log.BurnFees, "want", "1 burn fees (dst)")
 		}
+	} else {
```
favadi: all this logic should be moved to crawler package, storage should just store trade logs as is without additional logic.

tradelogs/crawler_v2.go
```diff
@@ -1,15 +1,27 @@
...
 	"github.com/ethereum/go-ethereum/core/types"
+	"go.uber.org/zap"
 )
 
+const (
+	//TradeExecute(address sender, address src, uint256 srcAmount, address destToken, uint256 destAmount, address destAddress)
+	tradeExecuteEvent = "0xea9415385bae08fe9f6dc457b02577166790cde83bb18cc340aac6cb81b824de"
+)
+
+func isZeroAddress(address ethereum.Address) bool {
```
favadi: this should be replaced with `blockchain.IsZeroAddress`?

tradelogs/crawler_v2.go
```diff
@@ -71,6 +137,16 @@ func (crawler *Crawler) assembleTradeLogsV2(eventLogs []types.Log) ([]common.Tra
 			if tradeLog.Timestamp, err = crawler.txTime.Resolve(log.BlockNumber); err != nil {
 				return nil, err
 			}
+			// when the tradelog does not contain burnfee and etherReceival event
+			// get tx receipt to get reserve address
+			if len(tradeLog.BurnFees) == 0 && blockchain.IsZeroAddress(tradeLog.SrcReserveAddress) {
```
favadi: please add a debug log here told that we are falling back to getting source/dest addresses.

tradelogs/crawler_test.go
```diff
@@ -58,7 +58,7 @@ func assertTradeLog(t *testing.T, tradeLog common.TradeLog) {
 }
 
 func TestCrawlerGetTradeLogs(t *testing.T) {
-	testutil.SkipExternal(t)
+	t.Skip("disable as this test require external resource")
```
cyberliem: Is the rebase overwritten a feature that included purposely?

tradelogs/crawler_v2.go
```diff
@@ -34,6 +42,60 @@ func (crawler *Crawler) fetchTradeLogV2(fromBlock, toBlock *big.Int, timeout tim
...
+		reserveAddr ethereum.Address
+	)
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+	receipt, err := crawler.ethClient.TransactionReceipt(ctx, txHash)
+	if err != nil {
+		return reserveAddr, err
+	}
+	for index := len(receipt.Logs) - 1; index >= 0; index-- {
+		if uint(index) > logIndex {
```
cyberliem: This seems to be unnecessary. We can simply loop from min(len(receipt.Logs), logIndex) to 0.

tradelogs/crawler_v2.go
```diff
@@ -71,6 +133,17 @@ func (crawler *Crawler) assembleTradeLogsV2(eventLogs []types.Log) ([]common.Tra
 			if tradeLog.Timestamp, err = crawler.txTime.Resolve(log.BlockNumber); err != nil {
 				return nil, err
 			}
+			// when the tradelog does not contain burnfee and etherReceival event
+			// get tx receipt to get reserve address
+			if len(tradeLog.BurnFees) == 0 && blockchain.IsZeroAddress(tradeLog.SrcReserveAddress) {
```
cyberliem: as mentioned before, walletFees contain reserveAddress. In this condition we should check for len(tradelog.WalletFee), if it ==0 then we process to get tx receipt, if it >0 we can get reserve Addr from here.

tradelogs/crawler_v2.go
```diff
@@ -71,6 +133,17 @@ func (crawler *Crawler) assembleTradeLogsV2(eventLogs []types.Log) ([]common.Tra
...
+				crawler.sugar.Debug("Tradelog have no burnfee and ethReceival event, fallback to get reserve address from tx receipt")
+				tradeLog.SrcReserveAddress, err = crawler.getTransactionReceipt(tradeLog.TransactionHash, 10*time.Second, log.Index)
+				if err != nil {
+					return nil, err
+				}
+			}
+			tradeLog = assembleTradeLogsReserveAddr(tradeLog, crawler.sugar)
```
cyberliem: Consider the case where len(burnFees)==0 and tradeLog.SrcReserveAddr isZero. We made the call to set tradeLog.SrcReserveAddress by getTransactionReceipt; but this function assembleTradeLogsReserveAddr() is called anyway. will it raise false warning of ["unexpected burn fees"](https://github.com/KyberNetwork/reserve-stats/pull/188/files#diff-590cafae26f0d44e2a6c576825f16d2aR93) ?

tradelogs/crawler_v2.go
halink0803: Got the idea. You should write this detail the first time instead of raising a question. Thanks

tradelogs/crawler_v2.go
halink0803: I don't think it will raise "unexpected burn fees" because getTransactionReceipt only update tradelog.SrcReserveAddr, not touching burnFees then it will not trigger the error. However, I will change to assemble  before getting the receipt.

tradelogs/crawler_test.go
halink0803: Yes.

tradelogs/crawler_v2.go
halink0803: Got the idea

tradelogs/crawler_v2.go
halink0803: Fixed in assembleTradeLogReserveAddr function.

tradelogs/crawler_v2.go
```diff
@@ -34,6 +44,64 @@ func (crawler *Crawler) fetchTradeLogV2(fromBlock, toBlock *big.Int, timeout tim
...
+		}
+	} else if blockchain.IsBurnable(log.DestAddress) {
+		if len(log.BurnFees) == 1 {
+			log.DstReserveAddress = log.BurnFees[0].ReserveAddress
+		} else {
+			sugar.Warnw("unexpected burn fees", "got", log.BurnFees, "want", "1 burn fees (dst)")
+		}
+	} else if len(log.WalletFees) != 0 {
+		if len(log.WalletFees) == 1 {
+			log.SrcReserveAddress = log.WalletFees[0].ReserveAddress
```
cyberliem: We need to check if this is src or dst reserve by checking the burnAble() condition. This might be a dst_reserve address

tradelogs/crawler.go
```diff
@@ -137,6 +137,7 @@ func fillEtherReceival(tradeLog common.TradeLog, logItem types.Log) (common.Trad
 	if err != nil {
 		return tradeLog, err
 	}
+	tradeLog.SrcReserveAddress = ethereum.BytesToAddress(logItem.Topics[1].Bytes())
```
cyberliem: Is it definitely SrcReserveAddress for DstReserveAddress? The logic of Src/Dst Reserve Address is binded to the logic of src_token and dst_token. I'm not quite sure if EtherReceival event always binded to SrcReserveAddress.

tradelogs/crawler_v2.go
halink0803: I just realized if we check burnable here then this logic will never reach. Because, if dest is burnable then burnFee slice should not be empty.

tradelogs/crawler.go
halink0803: DstReserve and SrcReserve now is base on which token is burnable, in this case it have no burnfee and the trade is from WETH to ETH, WETH and ETH is both not burnable, then I think the reserve address here can either be src or dest (or neither). I just choose it to be source reserve address.

PR #190, opened 2019-01-16
implements trade-logs-crawler planner

favadi:


tradelogs/cmd/trade-logs-crawler/main.go
```diff
@@ -1,15 +1,17 @@
...
 	"log"
 	"math"
 	"math/big"
 	"os"
 	"time"
 
+	"github.com/influxdata/influxdb/client/v2"
+	"github.com/urfave/cli"
+	"go.uber.org/zap"
+
```
halink0803: blank line

tradelogs/cmd/trade-logs-crawler/planner.go
```diff
@@ -0,0 +1,131 @@
...
+	"errors"
+	"io"
+	"math/big"
+	"time"
+
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/ethclient"
+	"github.com/urfave/cli"
+	"go.uber.org/zap"
+
```
halink0803: blank line

tradelogs/cmd/trade-logs-crawler/main.go
halink0803: Blank line

tradelogs/cmd/trade-logs-crawler/planner.go
halink0803: Blank line

tradelogs/cmd/trade-logs-crawler/main.go
```diff
@@ -231,71 +210,24 @@ func run(c *cli.Context) error {
...
-			}
-		}
-
-		if fromBlock.Cmp(toBlock) >= 0 {
-			return errors.New("fromBlock is bigger than toBlock")
+		var (
+			doneCh             = make(chan struct{})
+			fromBlock, toBlock *big.Int
+		)
+		if fromBlock, toBlock, err = planner.Next(); err == io.EOF {
```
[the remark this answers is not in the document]
favadi: io.EOF to indicate the end of data steam is a common pattern in golang.

- From [io/docs](https://godoc.org/io):

```
 EOF is the error returned by Read when no more input is available. Functions should return EOF only to signal a graceful end of input. If the EOF occurs unexpectedly in a structured data stream, the appropriate error is either ErrUnexpectedEOF or some other error giving more detail. 
```

- it is used in http [body reader](https://golang.org/pkg/net/http/)

```
        // For server requests the Request Body is always non-nil
        // but will return EOF immediately when no body is present.
        // The Server will close the request body. The ServeHTTP
        // Handler does not need to.
```

- [gzip](https://golang.org/pkg/compress/gzip/)

```
 Close closes the Reader. It does not close the underlying io.Reader. In order for the GZIP checksum to be verified, the reader must be fully consumed until the io.EOF. 
```

tradelogs/cmd/trade-logs-crawler/main.go
favadi: it is intentional to keep third party imports in a separate groups from internal improts.

tradelogs/cmd/trade-logs-crawler/planner.go
favadi: it is intentional to keep third party imports in a separate groups from internal improts.

tradelogs/cmd/trade-logs-crawler/main.go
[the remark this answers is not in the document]
cyberliem: In my opinion end of block is different from EOF, reuse it in this case might not be the best thing to do. Maybe we can define and export and err namely EOB and use it.

consider line 84 if storage engine change (bolt DB for example), where they read an incomplete record due to some reason and when it trying to read the buffer for a complete record, it meets io.EOF error, this program will return io.EOF but it's not end of blocks case.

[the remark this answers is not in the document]
halink0803: As @favadi once quote " EOF is the error returned by Read when no more input is available. Functions should return EOF only to signal a graceful end of input. If the EOF occurs unexpectedly in a structured data stream, the appropriate error is either ErrUnexpectedEOF or some other error giving more detail."

If storage engine change and it has errors, it should return UnexpectedEOF or some other details. So I think io.EOF here is reasonable.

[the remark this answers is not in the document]
cyberliem: Approved.

PR #192, opened 2019-01-18
2258 remove ETH_receival_sender

cyberliem:


tradelogs/common/types.go
```diff
@@ -30,8 +30,7 @@ type TradeLog struct {
 	BlockNumber     uint64        `json:"block_number"`
 	TransactionHash ethereum.Hash `json:"tx_hash"`
 
-	EtherReceivalSender ethereum.Address `json:"eth_receival_sender"`
-	EtherReceivalAmount *big.Int         `json:"eth_receival_amount"`
+	EtherReceivalAmount *big.Int `json:"eth_receival_amount"`
```
favadi: should this be called ETH amount? As we do not track EtherReceival event for V3 anymore.

PR #193, opened 2019-01-18
2252 enable test mode

cyberliem:


burnedfees/crawler/crawler_test.go
```diff
@@ -29,7 +31,13 @@ func (*mockStorage) LastBlock() (int64, error) {
...
 
 func TestBurnedFeesCrawlerExecute(t *testing.T) {
-	t.Skip("disable as this test require external resource")
+	testMode, err := testutil.GetTestMode()
+	if err != nil {
+		t.Skip("Can't get test mode. skip this external test")
+	}
+	if testMode != testutil.External {
+		t.Skip("disable as this test require external resource")
+	}
```
favadi: all of this should be grouped in a function like `testutil.SkipExternal(t *testing.T)`

PR #196, opened 2019-01-21
1702_untag_high_cardinals

cyberliem:


tradelogs/storage/influx.go
```diff
@@ -376,7 +369,12 @@ func (is *InfluxStorage) tradeLogToPoint(log common.TradeLog) ([]*client.Point,
 		logschema.DstAmount.String():  dstAmount,
 		logschema.EthUSDRate.String(): log.ETHUSDRate,
 
-		logschema.EthAmount.String(): ethAmount,
+		logschema.EthAmount.String():         ethAmount,
+		logschema.BlockNumber.String():       strconv.FormatUint(log.BlockNumber, 10),
```
favadi: we should store block number as int.

tradelogs/storage/influx.go
```diff
@@ -74,6 +72,7 @@ func (is *InfluxStorage) SaveTradeLogs(logs []common.TradeLog) error {
 	}
 
 	if err := is.influxClient.Write(bp); err != nil {
+		is.sugar.Errorw("saving error", "error", err)
```
favadi: there are some same `is.sugar` log message in this file, please add function name, params to its context.

tradelogs/storage/influx.go
```diff
@@ -290,6 +212,53 @@ func (is *InfluxStorage) queryDB(clnt client.Client, cmd string) (res []client.R
 	return res, nil
 }
 
+//setBurnFeeFields set all the dst/srcBurnFee amount fields
+func (is *InfluxStorage) setBurnFeeFields(log common.TradeLog, fields map[string]interface{}) error {
+	var logger = is.sugar.With(
+		"func", "tradelogs/storage/setBurnFeeTagsAndFields",
```
favadi: the func name does not match.

tradelogs/common/types.go
```diff
@@ -40,9 +40,15 @@ type TradeLog struct {
 	SrcAmount         *big.Int         `json:"src_amount"`
```
favadi: I notice that eth_amount tag name is `eth_receival_amount`, maybe fix it and update document?

tradelogs/storage/influx.go
```diff
@@ -290,6 +212,53 @@ func (is *InfluxStorage) queryDB(clnt client.Client, cmd string) (res []client.R
 	return res, nil
 }
 
+//setBurnFeeFields set all the dst/srcBurnFee amount fields
+func (is *InfluxStorage) setBurnFeeFields(log common.TradeLog, fields map[string]interface{}) error {
```
favadi: I feel like lot of logic in this function are unnecessary and duplicated in crawler somewhere. The logic should be simplifed to:

- if len BurnFees == 2 --> set both source burn fee and dest burn fee
- if len BurnFees == 1 and blockchain.IsBurnable(log.SrcAddress), set source burn fee, otherwise set dest burn fees

tradelogs/storage/influx.go
```diff
@@ -8,9 +8,7 @@ import (
 	"github.com/KyberNetwork/reserve-stats/lib/blockchain"
```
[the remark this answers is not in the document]
favadi: Also in this file

```go
	if log.SrcAddress == blockchain.ETHAddr {
		ethAmount = srcAmount
	} else if log.DestAddress == blockchain.ETHAddr {
		ethAmount = dstAmount
	} else {
		ethAmount = ethReceivalAmount
	}
```

should be remove, as ethReceivalAmount is really ethAmount now.

tradelogs/storage/influx.go
```diff
@@ -347,71 +314,37 @@ func (is *InfluxStorage) tradeLogToPoint(log common.TradeLog) ([]*client.Point,
...
-		}
-
 		amount, err := is.tokenAmountFormatter.FromWei(blockchain.KNCAddr, walletFee.Amount)
 		if err != nil {
 			return nil, err
 		}
-
-		fields := map[string]interface{}{
-			walletschema.Amount.String(): amount,
+		if walletFee.ReserveAddress == log.SrcReserveAddress {
```
favadi: this should be moved to a function like setBurnFees?

tradelogs/storage/influx.go
favadi: Anw, I'm not a fan of mutate parameter state in a function, please change the function signature to accept log and returns burn fee fields and merge it explicitly.

tradelogs/storage/influx_results.go
```diff
@@ -31,9 +27,15 @@ func (is *InfluxStorage) rowToAggregatedBurnFee(row []interface{}) (time.Time, f
 	if err != nil {
 		return ts, burnFee, reserve, err
 	}
-
-	reserve, err = influxdb.GetAddressFromInterface(row[2])
-
+	if (row[2] != nil) && (row[3] != nil) {
```
favadi: redundant ().

tradelogs/storage/influx_results.go
```diff
@@ -31,9 +27,15 @@ func (is *InfluxStorage) rowToAggregatedBurnFee(row []interface{}) (time.Time, f
...
+		panic("Logic fault : there should not be a record with both source and dest reserve address")
+	} else if row[2] != nil {
+		reserve, err = influxdb.GetAddressFromInterface(row[2])
+	} else if row[3] != nil {
+		reserve, err = influxdb.GetAddressFromInterface(row[3])
```
favadi: if there are both errors in line 33 and 35, the first one is ignored?

tradelogs/common/types.go
```diff
@@ -40,9 +40,15 @@ type TradeLog struct {
...
+	WalletAddress     ethereum.Address `json:"wallet_addr"`
 
-	BurnFees       []BurnFee   `json:"burn_fees"`
-	WalletFees     []WalletFee `json:"wallet_fees"`
+	SrcBurnAmount      float64 `json:"src_burn_amount"`
+	DstBurnAmount      float64 `json:"dst_burn_amount"`
+	SrcWalletFeeAmount float64 `json:"src_wallet_fee_amount"`
+	DstWalletFeeAmount float64 `json:"dst_wallet_fee_amount"`
+
+	BurnFees       []BurnFee   `json:"-"`
```
favadi: I took another look and all additional fields of `BurnFee` are discarded when saving/loading TradeLog, why not remove both `BurnFees` and `WalletFees` and fill SrcBurnAmount/DstBurnAmount/SrcWalletFeeAmount/DstWalletFeeAmount in crawler?

tradelogs/storage/wallet_fee.go
```diff
@@ -37,12 +37,13 @@ func (is *InfluxStorage) GetAggregatedWalletFee(reserveAddr, walletAddr, freq st
...
 	q := fmt.Sprintf(`
-		SELECT sum_amount FROM "%s"
-		WHERE reserve_addr = '%s' AND wallet_addr = '%s'
-		AND time >= '%s' AND time <= '%s' 
+		SELECT SUM(sum_amount) AS sum_amount FROM "%[1]s"
+		WHERE (src_rsv_addr = '%[2]s' OR dst_rsv_addr= '%[2]s') AND wallet_addr = '%[3]s'
+		AND time >= '%[4]s' AND time <= '%[5]s' GROUP BY time(1%[6]s) FILL(none) 
 	`, measurement, reserveAddr, walletAddr,
-		fromTime.UTC().Format(time.RFC3339), toTime.UTC().Format(time.RFC3339))
+		fromTime.UTC().Format(time.RFC3339), toTime.UTC().Format(time.RFC3339), freq)
```
favadi: why we add the GROUP BY in this new query?

tradelogs/common/types.go
cyberliem: Fixed to eth_amount

tradelogs/common/types.go
cyberliem: I did mentioned about this earlier, and the decision on this is to separate SrcReserveAddr and DstReservecAddr from the burnAmount v...v. In addition to that, burnAmount/FeeAmount etc... need tokenFormatter to change it to float type.

tradelogs/common/types.go
```diff
@@ -40,9 +40,15 @@ type TradeLog struct {
...
+	WalletFees     []WalletFee `json:"-"`
```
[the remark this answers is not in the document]
cyberliem: Hasn't been updated in a while, I did update it now.

tradelogs/storage/burnfee.go
```diff
@@ -30,8 +30,8 @@ func (is *InfluxStorage) GetAggregatedBurnFee(from, to time.Time, freq string, r
 		return nil, fmt.Errorf("invalid burn fee frequency %s", freq)
 	}
 
-	const queryTmpl = `SELECT sum_amount, reserve_addr FROM "{{.Measurement}}" WHERE '{{.From }}' <= time AND time <= '{{.To}}' ` +
-		`{{if len .Addrs}}AND ({{range $index, $element := .Addrs}}"reserve_addr" = '{{$element}}'{{if ne $index $.AddrsLastIndex}} OR {{end}}{{end}}){{end}}`
+	const queryTmpl = `SELECT sum_amount,src_rsv_addr,dst_rsv_addr FROM "{{.Measurement}}" WHERE '{{.From }}' <= time AND time <= '{{.To}}' AND (src_rsv_addr!='' OR dst_rsv_addr!='')` +
```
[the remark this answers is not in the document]
cyberliem: It's a safeguard, in case null record exist.

tradelogs/storage/burnfee.go
```diff
@@ -85,7 +85,7 @@ func (is *InfluxStorage) GetAggregatedBurnFee(from, to time.Time, freq string, r
 		if !ok {
 			result[reserve] = make(map[string]float64)
 		}
-		result[reserve][key] = amount
+		result[reserve][key] = result[reserve][key] + amount
```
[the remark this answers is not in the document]
cyberliem: add the comment.

tradelogs/storage/cq/country_cqs.go
```diff
@@ -107,10 +107,11 @@ func CreateCountryCqs(dbName string) ([]*libcq.ContinuousQuery, error) {
 		dbName,
 		dayResampleInterval,
 		dayResampleFor,
-		"SELECT SUM(amount) AS total_burn_fee INTO country_stats FROM burn_fees GROUP BY country",
+		"SELECT SUM(src_burn_fee)+SUM(dst_burn_fee) AS total_burn_fee INTO country_stats FROM trades GROUP BY country",
```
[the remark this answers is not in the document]
cyberliem: Will update the test.

tradelogs/storage/influx.go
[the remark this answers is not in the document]
cyberliem: removed.

tradelogs/storage/influx.go
cyberliem: changed.

tradelogs/storage/influx.go
```diff
@@ -210,54 +168,18 @@ func (is *InfluxStorage) LoadTradeLogs(from, to time.Time) ([]common.TradeLog, e
...
-			if !exist {
-				walletFeesByTxHash[txHash] = make(map[uint][]common.WalletFee)
-			}
-			walletFeesByTxHash[txHash][uint(tradeLogIndex)] = walletFees
-		}
-	}
-
 	// Get TradeLogs
-	if len(res[2].Series) == 0 {
+	if (len(res) == 0) || len(res[0].Series) == 0 || (len(res[0].Series[0].Values) == 0) {
```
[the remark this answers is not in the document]
cyberliem: change.d

tradelogs/storage/influx.go
cyberliem: The logic represented was tighten not to the len of BurnFees but to whether or not the dst/ src token is burnable ( extra log message is printout on this case) 

If there is an absolute need to change the old logic (depend on blockchain.BurnAble(token)) to new logic only concern number of burnFee, these log message will not be there anymore, abnormal cases might be overlook.

tradelogs/storage/influx.go
cyberliem: fixed

tradelogs/storage/influx.go
```diff
@@ -298,19 +267,17 @@ func (is *InfluxStorage) tradeLogToPoint(log common.TradeLog) ([]*client.Point,
 	}
 
 	tags := map[string]string{
-		logschema.BlockNumber.String(): strconv.FormatUint(log.BlockNumber, 10),
```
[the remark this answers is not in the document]
cyberliem: I overlook this cased. Fixed.

tradelogs/storage/influx.go
cyberliem: It's not that complicated so I decided not ot

tradelogs/storage/influx_results.go
cyberliem: fixed

tradelogs/storage/influx_results.go
cyberliem: return the err if occur

tradelogs/common/types.go
favadi: I still see `eth_receival_sender` field in _trade_logs.md file, please compare the real API response with the docs.

tradelogs/common/types.go
favadi: > I did mentioned about this earlier, and the decision on this is to separate SrcReserveAddr and DstReservecAddr from the burnAmount

Please remind me why we decide that.

> why not remove both `BurnFees` and `WalletFees` and fill SrcBurnAmount/DstBurnAmount/SrcWalletFeeAmount/DstWalletFeeAmount in crawler?

what is the downside of doing this?

tradelogs/storage/burnfee.go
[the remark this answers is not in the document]
favadi: Please remove this condition, we rather got a null record in response so we know something is wrong this the database.

tradelogs/storage/cq/country_cqs.go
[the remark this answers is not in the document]
favadi: seems to work fine now, closing.

tradelogs/storage/influx.go
favadi: Fair point, please keep the burn fees assertion.

tradelogs/storage/influx.go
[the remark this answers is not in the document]
favadi: confirmed that is fixed.

tradelogs/storage/burnfee.go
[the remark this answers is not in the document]
cyberliem: Removed.

tradelogs/storage/wallet_fee.go
cyberliem: I removed the GROUP BY Condition

tradelogs/common/types.go
cyberliem: > Please remind me why we decide that.

Since only the logic of crawler v2 to fill reserve_addr of the trade.  V3 does not come to that. We might have to intertwine these crawler logic to fill burnAmount for both v2 and v3

> what is the downside of doing this?

As mentioned, we need to include [token Formatter](https://github.com/KyberNetwork/reserve-stats/pull/196/files#diff-da88805979298949622417aa087cc5d5R231) to crawler to do that. In essence, we change crawler from only aware of big.Int to changing these number to float form.

tradelogs/storage/wallet_fee.go
cyberliem: I removed it and add in the logic to sum the two src/dst amount similar to burnFee

tradelogs/common/types.go
favadi: > As mentioned, we need to include [token Formatter](https://github.com/KyberNetwork/reserve-stats/pull/196/files#diff-da88805979298949622417aa087cc5d5R231) to crawler to do that. In essence, we change crawler from only aware of big.Int to changing these number to float form.

I see, we might need to use different structs for crawler/database presentation soon. But not in this task.

tradelogs/common/types.go
cyberliem: Updated

PR #203, opened 2019-01-29
2387 Add caching services

halink0803:


users/cacher/cacher.go
```diff
@@ -0,0 +1,166 @@
...
+	return nil
+}
+
+func (rc RedisCacher) cacheRichUser() error {
+	var (
+		logger = rc.sugar.With("func", "user/cacher/cachedUserInfo")
+	)
+	// read total trade 24h
+	query := fmt.Sprintf(`SELECT SUM(amount) as daily_fiat_amount FROM 
+	(SELECT eth_amount*eth_usd_rate as amount FROM trades WHERE time <= now() AND time >= (now()-24h) GROUP BY user_addr)
```
favadi: 1. time <= now() should be unnecessary as we can never write anything in future to trade logs database.

2. we should use tradelog.EthUSDRate and tradelog.EthAmount instead of hard code string.

3. I feel like the GROUP BY user_addr inside subquery is redundant, please check.

users/cacher/cacher.go
```diff
@@ -0,0 +1,166 @@
...
+		return err
+	}
+
+	// loop all user, check kyced
+	if len(res) == 0 || len(res[0].Series) == 0 || len(res[0].Series[0].Values) == 0 || len(res[0].Series[0].Values[0]) < 2 {
+		logger.Debugw("influx db is empty", "result", res)
+		return nil
+	}
+	kycedCap := common.NewUserCap(true)
+	nonKycedCap := common.NewUserCap(false)
```
favadi: these are initialized every time cacher run? Maybe moving it inside RedisCacher struct is more efficient?

users/cacher/cacher.go
```diff
@@ -0,0 +1,166 @@
...
+
+	for _, serie := range res[0].Series {
+		userAddress := serie.Tags["user_address"]
```
favadi: please use tradelog.UserAddress.

users/cacher/cacher.go
```diff
@@ -0,0 +1,166 @@
...
+		// check kyced
+		kyced, err := rc.isKyced(userAddress)
```
favadi: please document somewhere that this function needs to execute after cacheAllKycedUsers.

users/cacher/cacher.go
```diff
@@ -0,0 +1,166 @@
...
+			// if user is not rich then it is already cached before
+			continue
+		}
+		user := common.UserResponse{
+			Rich:  true,
+			KYCed: kyced,
+		}
+
+		// save to cache with 1 hour
+		rc.saveToCache(userAddress, user, expireTime)
```
favadi: the error is ignored.

users/cacher/cacher.go
favadi: Anw, I took a second look and think that we should cache the rich addresses in mem and lookup from there instead of written/read from redis just for this.

users/cacher/cacher.go
```diff
@@ -0,0 +1,166 @@
...
+		}
+
+		// check rich
+		userTradeAmount, err := influxdb.GetFloat64FromInterface(serie.Values[0][1])
+		if err != nil {
+			logger.Debugw("values second should be a float", "value", serie.Values[0][1])
+			return nil
+		}
+
+		if (kyced && userTradeAmount < kycedCap.DailyLimit) || (!kyced && userTradeAmount < nonKycedCap.DailyLimit) {
```
favadi: do we need this? I think maybe just always write to redis is simpler.

users/cacher/cacher.go
```diff
@@ -0,0 +1,166 @@
...
+			KYCed: kyced,
+		}
+
+		// save to cache with 1 hour
+		rc.saveToCache(userAddress, user, expireTime)
+	}
+	return nil
+}
+
+func (rc RedisCacher) saveToCache(key string, value common.UserResponse, expireTime time.Duration) error {
```
favadi: I think we should use redis pipeline, prepare the commands and do one roundtrip to the server instead of multiple times.

users/cacher/cacher.go
```diff
@@ -0,0 +1,166 @@
...
+	data, err := json.Marshal(value)
```
favadi: I think we should use 2 separate prefixes instead of doing this.

- SET kyced:0x1234589 1 (no expired time)
- SET rich:0x123456789 1 1hour

So if somehow the cacher service crashed, the KYCed still works.

users/cmd/users-public-cacher/main.go
```diff
@@ -0,0 +1,77 @@
...
+	redisCacheClient, err := rediscache.NewClientFromContext(c)
+	if err != nil {
+		return err
+	}
+
+	sugar.Debugw("Initiated redis cached", "cache", redisCacheClient)
+
+	redisCacher := cacher.NewRedisCacher(sugar, userDB, influxDBClient, redisCacheClient)
+
+	redisCacher.CacheUserInfo()
```
favadi: the error is ignored.

users/cmd/users-public-stats/main.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/KyberNetwork/reserve-stats/lib/app"
+	"github.com/KyberNetwork/reserve-stats/lib/httputil"
+	_ "github.com/KyberNetwork/reserve-stats/lib/httputil/validators" // import custom validator functions
+	rediscache "github.com/KyberNetwork/reserve-stats/lib/redis"
+	server "github.com/KyberNetwork/reserve-stats/users/public-server"
+	"github.com/KyberNetwork/tokenrate/coingecko"
+)
+
+func main() {
+	app := app.NewApp()
```
favadi: app variable overrided app package.

users/cmd/users-public-stats/main.go
```diff
@@ -0,0 +1,55 @@
...
+import (
+	"log"
+	"os"
+
+	"github.com/urfave/cli"
+
+	"github.com/KyberNetwork/reserve-stats/lib/app"
+	"github.com/KyberNetwork/reserve-stats/lib/httputil"
+	_ "github.com/KyberNetwork/reserve-stats/lib/httputil/validators" // import custom validator functions
+	rediscache "github.com/KyberNetwork/reserve-stats/lib/redis"
```
favadi: libredis is better alias.

users/cmd/users-public-stats/main.go
```diff
@@ -0,0 +1,55 @@
...
+	logger, err := app.NewLogger(c)
+	if err != nil {
+		return err
+	}
+	defer logger.Sync()
+
+	sugar := logger.Sugar()
+	sugar.Info("Run user stats public service")
+
+	cachedClient, err := rediscache.NewClientFromContext(c)
```
favadi: just call it redisClient is clearer.

users/http/cached_rate_provider.go
```diff
@@ -8,15 +8,17 @@ import (
 	"go.uber.org/zap"
 )
 
-func newCachedRateProvider(sugar *zap.SugaredLogger, provider tokenrate.ETHUSDRateProvider, timeout time.Duration) *cachedRateProvider {
-	return &cachedRateProvider{
+//NewCachedRateProvider return cached provider for eth usd rate
+func NewCachedRateProvider(sugar *zap.SugaredLogger, provider tokenrate.ETHUSDRateProvider, timeout time.Duration) *CachedRateProvider {
```
favadi: this should be moved to lib/somewhere? It does not make sense users-public-stats includes users/http package.

users/cacher/cacher.go
halink0803: This logic check if user is not rich, then we do not need to write anymore, save many write duplicate data.

users/cacher/cacher.go
halink0803: Agree

users/http/cached_rate_provider.go
halink0803: Already move to lib/httputil as this cachedRateProvider only use in user server.

lib/redis/flags.go
```diff
@@ -0,0 +1,56 @@
...
+func NewClientFromContext(c *cli.Context) (*redis.Client, error) {
+	redisEndpoint := c.String(redisEndpointFlag)
+	redisPassword := c.String(redisPasswordFlag)
+
+	if err := validation.Validate(redisEndpoint, validation.Required, is.URL); err != nil {
+		return nil, err
+	}
+	redisClient := redis.NewClient(&redis.Options{
+		Addr:     redisEndpoint,
+		Password: redisPassword,
```
favadi: database flag is missing?

users/cacher/cacher.go
```diff
@@ -0,0 +1,188 @@
...
+		return err
+	}
+	logger.Debugw("addresses from postgres", "addresses", addresses)
+
+	pipe := rc.redisClient.Pipeline()
+	defer func() {
+		if _, err := pipe.Exec(); err != nil {
+			logger.Errorw("pipeline exec error", "error", err.Error())
+		}
+	}()
```
favadi: the error is ignored, `pipe.Exec` failed. So `pipe.Discard/Close` if `saveToCache` failed.

users/cacher/cacher.go
```diff
@@ -0,0 +1,188 @@
...
+		// save to cache with 1 hour
+		if err := rc.saveToCache(pipe, fmt.Sprintf("%s:%s", richPrefix, userAddress), user, expireTime); err != nil {
+			return err
+		}
+	}
+
+	return err
+}
+
+func (rc *RedisCacher) saveToCache(pipeline redis.Pipeliner, key string, value common.UserResponse, expireTime time.Duration) error {
```
favadi: saveToCache is not not a good name as the function only push the commands to pipeline.

users/cacher/cacher.go
```diff
@@ -0,0 +1,188 @@
...
+}
+
+func (rc *RedisCacher) cacheRichUser() error {
+	var (
+		logger = rc.sugar.With("func", "user/cacher/cacheRichUser")
+	)
+	// read total trade 24h
+	query := fmt.Sprintf(`SELECT SUM(amount) as daily_fiat_amount FROM 
+	(SELECT %s*%s as amount FROM trades WHERE time >= (now()-24h))
+	GROUP BY user_addr`, logSchema.EthAmount.String(), logSchema.EthUSDRate.String())
```
favadi: we can optimize the query by SELECT SUM > nonKYCed cap to returns fewer records.

users/cacher/cacher.go
favadi: if you decided to go with storing everything to store one record, please use HMSET instead of json encoding/decosing.

favadi:
The data is duplciated in both kyced and rich namespaces, please just us SET/GET with 2 namespaces instead of hash with redundant data.

PR #213, opened 2019-02-20
2561 fix wrong wallet fee with new db implementation

cyberliem:


tradelogs/storage/influx.go
```diff
@@ -245,11 +245,35 @@ func (is *InfluxStorage) getBurnAmount(log common.TradeLog) (float64, float64, e
...
+		if walletFee.ReserveAddress == log.SrcReserveAddress && !srcAmountSet {
+			srcAmount = amount
+			srcAmountSet = true
+		} else if walletFee.ReserveAddress == log.DstReserveAddress {
+			dstAmount = amount
+		} else {
+			logger.Warnw("unexpected wallet fees with unrecognized reserve address", "wallet fee", walletFee)
+		}
+	}
+	return srcAmount, dstAmount, nil
```
favadi: could this function be simplified to:

- if len wallet_fees ==2 {return src=wallet_fees[0], dst=wallet_fees[1], nil}
- if len(wallet_fees==1) {return src=wallet_fees[0]}, 0, nil}
- if len(wallet_fees==0} {return 0, 0, nil}
- else panic

cyberliem: this bound to the logic of src/dst burn able. Abnormality case must came with WARN log.

Inaddition . If we based it on len(wallet_fees), when len(wallet_fees)==1, the amount might belong to different reserve_addr. Since our API return burn fee on reserve_addr and wallet_addr, we must separate src and dst wallet fee amount

favadi: You are right, this is fine.

PR #223, opened 2019-02-25
2519 Add accounting listed token fetcher

halink0803:


lib/blockchain/token_symbol.go
```diff
@@ -102,6 +112,51 @@ func (t *TokenSymbol) Symbol(address common.Address) (string, error) {
...
+		err  error
+	)
+	if val, ok := t.cachedName.Load(address); ok {
+		if name, ok = val.(string); !ok {
+			return "", errors.New("invalid token name stored in cached symbol")
+		}
+		return name, nil
+	}
+	name, err = getName1(address, t.ethClient)
+	if err != nil && strings.Contains(err.Error(), "abi: cannot marshal") {
```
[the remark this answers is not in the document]
cyberliem: In addition to that, the logic of deciding whether this is a ERC20 or ERC20Type 2 can be made without us concerning about the type of error return.

Why don't we call both function parallely, and wait() for the result, picking any result without error.  If both return errors, we return a combination of errors.  Please discuss with @ductm54 for more information since he was the original author of this package.

users/migration/migrate.go
```diff
@@ -8,6 +8,7 @@ import (
 
 	"github.com/KyberNetwork/reserve-stats/lib/boltutil"
 	"github.com/KyberNetwork/reserve-stats/lib/pgsql"
+
```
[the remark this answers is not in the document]
halink0803: separate internal package and external package

accounting/cmd/accounting-listed-token-fetcher/main.go
```diff
@@ -0,0 +1,133 @@
...
+		if err != nil {
+			return err
+		}
+		result = append(result, common.ListedToken{
+			Address: v.Hex(),
+			Symbol:  symbol,
+			Name:    name,
+		})
+	}
+	resultJSON, _ := json.Marshal(result)
```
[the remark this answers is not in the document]
halink0803: just for print out to cli.

lib/blockchain/token_symbol.go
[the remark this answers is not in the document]
halink0803: decided to try to call ERC20Type2 whenever error happen, combine error if they happen. Not call them parallel because it is not necessary, and save some request.

users/migration/migrate.go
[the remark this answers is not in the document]
favadi: it is a minor issue, but please do not commit unrelated changes next time.

lib/blockchain/token_symbol.go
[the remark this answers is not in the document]
favadi: it is OK to call both requests concurrently as most of the token will return with type1 already, type2 tokens are minor.

accounting/storage/listed_token_storage_postgres.go
```diff
@@ -0,0 +1,78 @@
...
+		db:    db,
+	}, nil
+}
+
+//CreateOrUpdate add or edit an record in the tokens table
+func (ltd *ListedTokenDB) CreateOrUpdate(tokens []common.ListedToken) error {
+	var (
+		logger = ltd.sugar.With("func", "accounting/storage.CreateOrUpdate")
+	)
+	logger.Info("create or upate token listed")
```
cyberliem: upate ->update

accounting/storage/listed_token_storage_postgres.go
```diff
@@ -0,0 +1,78 @@
...
+	tx, err := ltd.db.Beginx()
+	if err != nil {
+		return err
+	}
+	defer pgsql.CommitOrRollback(tx, logger, &err)
+	for _, token := range tokens {
+		// check if token already exist in db
```
cyberliem: It is possible to use ON CONFLICT ON CONSTRAINT to define this behaviour instead of doing a db read for this

lib/blockchain/token_symbol.go
```diff
@@ -102,6 +112,66 @@ func (t *TokenSymbol) Symbol(address common.Address) (string, error) {
...
+			return "", errors.New("invalid token name stored in cached symbol")
+		}
+		return result, nil
+	}
+	wg := sync.WaitGroup{}
+	for _, getNameFn := range getNameFns {
+		wg.Add(1)
+		go func(fn func(common.Address, bind.ContractBackend) (string, error)) {
+			defer wg.Done()
+			name, err = fn(address, t.ethClient)
```
favadi: we do nothing with the err?

accounting/cmd/accounting-listed-token-fetcher/main.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+	tokenSymbol, err := blockchain.NewTokenSymbolFromContext(c)
+	if err != nil {
+		return err
+	}
+
+	return getListedToken(ethClient, block, reserveAddr, tokenSymbol, sugar)
+}
+
+func getListedToken(ethClient *ethclient.Client, block *big.Int, reserveAddr ethereum.Address,
```
favadi: this should be move out of main.go file.

accounting/cmd/accounting-listed-token-fetcher/main.go
```diff
@@ -0,0 +1,137 @@
...
+	conversionRateContractClient, err := contracts.NewConversionRates(conversionRatesContract, ethClient)
+	if err != nil {
+		return err
+	}
+	listedTokens, err := conversionRateContractClient.GetListedTokens(callOpts)
+	if err != nil {
+		return err
+	}
+	for _, v := range listedTokens {
+		symbol, err := tokenSymbol.Symbol(v)
```
favadi: the err is not shadowed?

accounting/cmd/accounting-listed-token-fetcher/main.go
```diff
@@ -0,0 +1,137 @@
...
+		if err != nil {
+			return err
+		}
+		name, err := tokenSymbol.Name(v)
```
favadi: the err is not shadowed?

lib/blockchain/token_symbol.go
```diff
@@ -102,6 +112,66 @@ func (t *TokenSymbol) Symbol(address common.Address) (string, error) {
 	return symbol, nil
 }
 
+var getNameFns = []func(common.Address, bind.ContractBackend) (string, error){
+	getName1,
+	getName2,
+}
+
+//Name return name of token
+func (t *TokenSymbol) Name(address common.Address) (string, error) {
```
favadi: this struct now returns more than just symbol, maybe rename it to `TokenInfo` or something else?

lib/blockchain/token_symbol.go
favadi: this might cause data race as name, err are both changed by different go routines.

accounting/listed-token-fetcher/fetcher.go
```diff
@@ -0,0 +1,117 @@
...
+	sugar *zap.SugaredLogger) *Fetcher {
+	return &Fetcher{
+		ethClient:                 ethClient,
+		contractTimestampResolver: contractTimestampResolver,
+		sugar:                     sugar,
+	}
+}
+
+func updateListedToken(listedToken []common.ListedToken, symbol, name string, address ethereum.Address, timestamp time.Time) []common.ListedToken {
+	timestampms := timeutil.TimeToTimestampMs(timestamp)
```
favadi: timestampms --> timestamps

accounting/listed-token-fetcher/fetcher.go
```diff
@@ -0,0 +1,117 @@
...
+	for _, token := range listedToken {
+		if token.Symbol == symbol {
```
favadi: IIRC the specs said both symbol, name should matched to be considered 2 tokens are the same.

accounting/listed-token-fetcher/fetcher.go
```diff
@@ -0,0 +1,117 @@
...
+	return &Fetcher{
+		ethClient:                 ethClient,
+		contractTimestampResolver: contractTimestampResolver,
+		sugar:                     sugar,
+	}
+}
+
+func updateListedToken(listedToken []common.ListedToken, symbol, name string, address ethereum.Address, timestamp time.Time) []common.ListedToken {
+	timestampms := timeutil.TimeToTimestampMs(timestamp)
+	for _, token := range listedToken {
```
favadi: is there anyway to store the data in a map to avoid expensive looping every time?

PR #224, opened 2019-02-25
External service clients

halink0803:


lib/externalclient/binance_client.go
```diff
@@ -0,0 +1,195 @@
...
+		return respBody, err
+	}
+	defer func() {
+		if cErr := resp.Body.Close(); cErr != nil {
+			logger.Errorw("Response body close error", "error", cErr.Error())
+		}
+	}()
+	switch resp.StatusCode {
+	case 429:
+		err = errors.New("breaking binance request rate limit")
```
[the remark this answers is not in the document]
halink0803: Yes, binance return error code 429: https://github.com/binance-exchange/binance-official-api-docs/blob/master/rest-api.md

lib/externalclient/binance_client.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+	req.Header.Add("Accept", "application/json")
+	if signNeeded {
+		q := req.URL.Query()
+		sig := url.Values{}
+		req.Header.Set("X-MBX-APIKEY", bc.APIKey)
+		q.Set("timestamp", fmt.Sprintf("%d", int64(timepoint)))
+		q.Set("recvWindow", "5000")
+		sig.Set("signature", bc.Sign(q.Encode()))
+		req.URL.RawQuery = q.Encode() + "&" + sig.Encode()
```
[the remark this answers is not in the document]
halink0803: values and signature must be at the end, otherwise the signature will be not valid.

lib/externalclient/huobi_client.go
```diff
@@ -0,0 +1,204 @@
...
+func NewHuobiClient(apiKey, secretKey string, sugar *zap.SugaredLogger) *HuobiClient {
+	return &HuobiClient{
+		APIKey:    apiKey,
+		SecretKey: secretKey,
+		sugar:     sugar,
+	}
+}
+
+//Sign return sign of request
+func (hc *HuobiClient) Sign(msg string) string {
```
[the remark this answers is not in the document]
halink0803: No, try to replace one with the other it does not work.

lib/externalclient/binance_client.go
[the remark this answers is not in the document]
favadi: > values and signature must be at the end, otherwise the signature will be not valid.

I did a test and it is true, please add a comment.

Also, when doing the test, I found out binance still returns 200 if the signature parameter is not at the end of the query string.

> 2019/02/27 10:40:51 {"msg":"Signature for this request is not valid.","success":false}
> 2019-02-27T10:40:51.999+0700	INFO	binance/binance_client_test.go:34	{map[]}

lib/externalclient/huobi/huobi_client.go
```diff
@@ -0,0 +1,200 @@
...
+	if signNeeded {
+		timestamp := fmt.Sprintf("%s", time.Now().UTC().Format("2006-01-02T15:04:05"))
+		params["SignatureMethod"] = "HmacSHA256"
+		params["SignatureVersion"] = "2"
+		params["AccessKeyId"] = hc.APIKey
+		params["Timestamp"] = timestamp
+	}
+	var sortedParams []string
+	for k := range params {
+		sortedParams = append(sortedParams, k)
```
favadi: do we really need to sort params here? I think q.Encode() supposed to sort for us already.

lib/externalclient/huobi/huobi_client.go
```diff
@@ -0,0 +1,200 @@
...
+	if signNeeded {
+		q := req.URL.Query()
+		method := req.Method
+		auth := q.Encode()
+		hostname := req.URL.Hostname()
+		path := req.URL.Path
+		payload := strings.Join([]string{method, hostname, path, auth}, "\n")
+		signature, err := hc.sign(payload)
+		if err != nil {
+			log.Fatal(err)
```
favadi: please return the error.

lib/externalclient/huobi/huobi_client.go
```diff
@@ -0,0 +1,200 @@
...
+	case 404:
+		err = errors.New("api not found")
+		break
+	case 429:
+		err = errors.New("breaking Huobi request rate limit")
+		break
+	case 500:
+		err = errors.New("500 from Huobi, its fault")
+		break
+	case 200:
```
favadi: please check if we need to check "success": true/false like binance above.

lib/externalclient/huobi/huobi_client.go
```diff
@@ -0,0 +1,200 @@
...
+	)
+	endpoint := fmt.Sprintf("%s/v1/order/orders", HuobiEndpoint)
+	res, err := hc.sendRequest(
+		http.MethodGet,
+		endpoint,
+		map[string]string{
+			"states":     "filled",
+			"symbol":     strings.ToLower(symbol),
+			"start-date": startDate.Format("2016-06-09"),
+			"endDate":    endDate.Format("2018-03-08"),
```
favadi: the format layout should be pass as reference to `Mon Jan 2 15:04:05 -0700 MST 2006`.

lib/externalclient/binance/binance_client.go
```diff
@@ -0,0 +1,195 @@
...
+		if err = json.NewDecoder(resp.Body).Decode(&response); err != nil {
+			break
+		}
+		err = fmt.Errorf("binance return with code: %d - %s", resp.StatusCode, response.Msg)
+	}
+	return respBody, err
+}
+
+//GetTradeHistory return history of trading on binance
+func (bc *Client) GetTradeHistory(symbol string, fromID int64) ([]TradeHistory, error) {
```
favadi: this function should accept from time/to time? how would user know last id to pass to this?

lib/externalclient/huobi/huobi_client_test.go
```diff
@@ -0,0 +1,48 @@
...
+	testutil.SkipExternal(t)
+	logger, err := zap.NewDevelopment()
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer logger.Sync()
+	sugar := logger.Sugar()
+
+	huobiAPIKey, ok := os.LookupEnv("HUOBI_API_KEY")
+	assert.True(t, ok, "huobi api key is not set")
```
favadi: should use `require.True`.

lib/externalclient/huobi/huobi_client_test.go
```diff
@@ -0,0 +1,48 @@
...
+
+	huobiSecretKey, ok := os.LookupEnv("HUOBI_SECRET_KEY")
+	assert.True(t, ok, "huobi secret key is not set")
```
favadi: should use `require.True`.

lib/externalclient/huobi/huobi_client_test.go
```diff
@@ -0,0 +1,48 @@
...
+
+	huobiClient := NewHuobiClient(huobiAPIKey, huobiSecretKey, sugar)
+	accounts, err := huobiClient.GetAccounts()
+	assert.Nil(t, err, fmt.Sprintf("get account fee error: %s", err))
```
favadi: use `assert.Error`.

lib/externalclient/huobi/huobi_client_test.go
```diff
@@ -0,0 +1,48 @@
...
+	assert.NotNil(t, accounts, "get accounts nil")
```
favadi: should use `assert.Empty`.

lib/externalclient/huobi/huobi_client_test.go
```diff
@@ -0,0 +1,48 @@
...
+
+	//fixed timestamp for test
+	startDate := time.Date(2018, time.January, 1, 0, 0, 0, 0, time.UTC)
+	endDate := time.Date(2019, time.February, 1, 0, 0, 0, 0, time.UTC)
+
+	tradeHistory, err := huobiClient.GetTradeHistory("bixeth", startDate, endDate)
+	assert.Nil(t, err, fmt.Sprintf("get history error: %v", err))
+	assert.NotNil(t, tradeHistory, "trade history is nil")
```
favadi: same comment as above.

lib/externalclient/huobi/huobi_client_test.go
```diff
@@ -0,0 +1,48 @@
...
+	sugar.Info(tradeHistory)
+
+	withdrawHistory, err := huobiClient.GetWithdrawHistory()
+	assert.Nil(t, err)
+	assert.NotNil(t, withdrawHistory)
```
favadi: same comment as above.

lib/externalclient/huobi/interface.go
```diff
@@ -0,0 +1,9 @@
+package huobi
+
+import "time"
+
+//Interface is for huobi api client
+type Interface interface {
+	GetTradeHistory(symbol string, startDate, endDate time.Time) (TradeHistoryList, error)
+	GetWithdrawHistory() (WithdrawHistoryList, error)
```
favadi: this method should accept from time to time?

lib/externalclient/huobi/huobi_client.go
halink0803: Create a new task to handle this situation later: 2637

lib/externalclient/binance/binance_client.go
halink0803: First time we can input fromID as 0, later query we should know the last id saved.

lib/externalclient/binance_client.go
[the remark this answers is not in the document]
halink0803: Commented and also add the task to handle this situation: 2637

PR #228, opened 2019-02-28
2524 implement rate limiter 

cyberliem:


lib/etherscan/etherscan_client_test.go
```diff
@@ -37,3 +40,39 @@ func TestEtherScanClient(t *testing.T) {
 	assert.NoError(t, err, fmt.Sprintf("get normal tx error: %v", err))
 	assert.NotEmpty(t, normalTxs, "get normal txs failed")
 }
+
+func TestEtherScanClientWithRateLimiter(t *testing.T) {
+	//commented out this and run with different rps to see its behaviour
```
favadi: if this test need to be observed by human eyes, it should be always disabled and only manually enable by developer when need. Please replace testutil.SkipExternal with t.Skip().

lib/etherscan/etherscan_client_test.go
```diff
@@ -37,3 +40,39 @@ func TestEtherScanClient(t *testing.T) {
...
+	etherscanClient := etherscan.New(etherscan.Network("api"), etherscanAPIKey)
+	limiter := rate.NewLimiter(rate.Limit(float64(rps)), 1)
+	etherscanClient.BeforeRequest = limitRate(limiter)
+
+	for i := 0; i < 20; i++ {
+		wg.Add(1)
+		go func(index int) {
+			defer wg.Done()
+			_, err := etherscanClient.EtherTotalSupply()
+			sugar.Debugw("finshied a request", "request_number", index, "finish_time", time.Now(), "error", err)
```
favadi: finshied, typo.

lib/binance/binance_client.go
```diff
@@ -24,17 +26,19 @@ const (
...
-	sugar     *zap.SugaredLogger
+	APIKey      string
+	SecretKey   string
+	sugar       *zap.SugaredLogger
+	rateLimiter *rate.Limiter
 }
 
 //NewBinance return a new client for binance api
-func NewBinance(apiKey, secretKey string, sugar *zap.SugaredLogger) *Client {
+func NewBinance(apiKey, secretKey string, sugar *zap.SugaredLogger, limiter *rate.Limiter) *Client {
```
favadi: We should have an interface for limiter, maybe

```go
type binanceLimiter interface {
  WaitN(n uint64)
}
```
And the limiter should be a an optional configuration with default request per second taken from `https://api.binance.com/api/v1/exchangeInfo`.

lib/huobi/huobi_client.go
```diff
@@ -25,17 +27,19 @@ const (
...
-	sugar     *zap.SugaredLogger
+	APIKey      string
+	SecretKey   string
+	sugar       *zap.SugaredLogger
+	rateLimiter *rate.Limiter
 }
 
 //NewClient return a new HuobiClient instance
-func NewClient(apiKey, secretKey string, sugar *zap.SugaredLogger) *Client {
+func NewClient(apiKey, secretKey string, sugar *zap.SugaredLogger, limiter *rate.Limiter) *Client {
```
favadi: same comment as binance client.

cyberliem: changed to interface

lib/binance/binance_client.go
cyberliem: changed to . interface

PR #230, opened 2019-03-01
2534 Accounting binance trades fetcher

halink0803:


accounting/cmd/accounting-binance-trades-fetcher/main.go
```diff
@@ -0,0 +1,137 @@
...
+	)
+
+	app.Flags = append(app.Flags, binance.NewCliFlags()...)
+
+	if err := app.Run(os.Args); err != nil {
+		log.Fatal(err)
+	}
+}
+
+func getTradeHistoryForOneSymBol(binanceClient *binance.Client, fromTime, toTime time.Time, symbol string,
```
cyberliem: I suggest that we create a Fetcher and write a wrapper function for this. Maybe something like
```go
type Fetcher struct {
   sugar: *zap.SugaredLogger
   client: *binance.Client
   storage: *binance.Storage
}
```

and wrap the getTradeHistory as `Fetcher.GetTradeHistory()` with the control logic and in the main program only call this fetcher.Run.

Why? 
-  More control on fetching jobs,  later we will need to make this into a daemon mode. 
-  Align the fetching/ storage operations. 

Please advise @favadi

favadi: @cyberliem has a point, the logic should be moved out of main.go to a separate package.

halink0803: I got the idea. It is a better approach, will change like that.

halink0803:
@cyberliem updated PR with rate limiter 20rps and burst size is 5 and it seems working.

accounting/binance-fetcher/fetcher.go
```diff
@@ -0,0 +1,85 @@
...
+		client: client,
+	}
+}
+
+func (f *Fetcher) getTradeHistoryForOneSymBol(fromTime, toTime time.Time, symbol string,
+	tradeHistories *sync.Map, wg *sync.WaitGroup) error {
+	var (
+		logger = f.sugar.With("func", "accounting/binance-fetcher.getTradeHistoryForOneSymbol")
+	)
+	result := []binance.TradeHistory{}
```
cyberliem: should move all the declaration to var ( ...  ) for better readability

accounting/cmd/accounting-binance-trades-fetcher/main.go
```diff
@@ -0,0 +1,83 @@
...
+
+	sugar := logger.Sugar()
+	sugar.Info("initiate fetcher")
+
+	binanceClient, err := binance.NewClientFromContext(c, sugar)
+	if err != nil {
+		return err
+	}
+
+	if c.Uint64(fromFlag) != 0 {
```
cyberliem: Do we have a handler for when fromFlag is 0? Or should we return error if fromFlag =0?

halink0803: Nope, fromFlag and toFlag is both optional, if fromFlag is 0 then fetcher will get all trades.

accounting/binance-fetcher/fetcher.go
halink0803: Yes

accounting/binance-fetcher/fetcher.go
```diff
@@ -0,0 +1,85 @@
...
+		logger         = f.sugar.With("func", "accounting/binance-fetcher.getTradeHistory")
+	)
+	// get list token
+	exchangeInfo, err := f.client.GetExchangeInfo()
+	if err != nil {
+		return err
+	}
+	tokenPairs := exchangeInfo.Symbols
+	wg := sync.WaitGroup{}
+	for _, pair := range tokenPairs {
```
favadi: there are like 500 token pairs, maybe we should process in small batch and sleep, retry if got 429, 418, 403?

accounting/binance-fetcher/fetcher.go
```diff
@@ -68,12 +94,18 @@ func (f *Fetcher) GetTradeHistory(fromTime, toTime time.Time) error {
...
 	wg := sync.WaitGroup{}
-	for _, pair := range tokenPairs {
-		wg.Add(1)
-		logger.Debugw("token", "pair", pair.Symbol)
-		go f.getTradeHistoryForOneSymBol(fromTime, toTime, pair.Symbol, &tradeHistories, &wg)
+	index := 0
+	for index < len(tokenPairs) {
+		for count := 0; count < batchSize && index+count < len(tokenPairs); count++ {
+			pair := tokenPairs[index+count]
+			log.Println(index + count)
```
cyberliem: remove this log or use proper logger for it

accounting/binance-fetcher/fetcher.go
```diff
@@ -68,12 +94,18 @@ func (f *Fetcher) GetTradeHistory(fromTime, toTime time.Time) error {
...
+			wg.Add(1)
+			logger.Debugw("token", "pair", pair.Symbol)
+			go f.getTradeHistoryForOneSymBol(fromTime, toTime, pair.Symbol, &tradeHistories, &wg)
```
cyberliem: It is possible to write an anonymous function similar to this:
```go
go func() {
  defer wg.Done()
  data,err:=f.getTradeHistoryForOneSymBol(fromTime, toTime, pair.Symbol)
  //save data
}()
```
to avoid passing map/ waitgroup around. This is just an suggestion for easier readability, not a must change. According to @favadi , this will helps determine the flow of data in program.

cyberliem: in addition to that, the error return from ```getTradeHistoryForOneSymbol``` is not asserted. This is the reason why errorGroup might be better for this implementation.

accounting/binance-fetcher/fetcher.go
```diff
@@ -75,36 +71,44 @@ func (f *Fetcher) getTradeHistoryForOneSymBol(fromTime, toTime time.Time, symbol
...
+			errGroup.Go(
+				func(pair binance.Symbol) func() error {
+					return func() error {
+						logger.Debugw("token", "pair", pair.Symbol)
+						oneSymbolTradeHistory, err := f.getTradeHistoryForOneSymBol(fromTime, toTime, pair.Symbol)
+						if err != nil {
+							return err
+						}
+						tradeHistories.Store(pair.Symbol, oneSymbolTradeHistory)
+						return nil
```
cyberliem: should the len of tradeHistory is asserted to be not 0 before storing?

halink0803: Yes, reasonable.

accounting/binance-fetcher/fetcher.go
```diff
@@ -0,0 +1,123 @@
...
+						if len(oneSymbolTradeHistory) != 0 {
+							tradeHistories.Store(pair.Symbol, oneSymbolTradeHistory)
+						}
+						return nil
+					}
+				}(pair),
+			)
+		}
+		if err := errGroup.Wait(); err != nil {
+			return err
```
favadi: we don't need to handle 429 error? I think binance will ban our IP if we attempt for request too many times after receiving 429?

accounting/cmd/accounting-binance-trades-fetcher/main.go
```diff
@@ -0,0 +1,110 @@
...
+		cli.Uint64Flag{
+			Name:   fromFlag,
+			Usage:  "From timestamp(millisecond) to get trade history from",
+			EnvVar: "FROM",
+		},
+		cli.Uint64Flag{
+			Name:   toFlag,
+			Usage:  "To timestamp(millisecond) to get trade history to",
+			EnvVar: "TO",
+		},
```
favadi: please use the from/to millis flags in lib/timeutil/flags.go.

accounting/cmd/accounting-binance-trades-fetcher/main.go
favadi: Binance API has the ability to fetch all trades when startTime = 0? If yes, please keep default from time == 0 in this task and ignore my comment in the binance withdrawal history task.

accounting/binance-fetcher/fetcher.go
```diff
@@ -0,0 +1,123 @@
...
+		logger.Debugw("trade history for", "symbol", symbol, "history", tradeHistoriesResponse)
+		result = append(result, tradeHistoriesResponse...)
+		lastTrade := tradeHistoriesResponse[len(tradeHistoriesResponse)-1]
+		startTime = timeutil.TimestampMsToTime(lastTrade.Time + 1)
+	}
+	return result, nil
+}
+
+//GetTradeHistory get all trade history from trades for all token
+func (f *Fetcher) GetTradeHistory(fromTime, toTime time.Time) error {
```
favadi: the fetcher should returns the trade history, right? I only it returns an error.

accounting/cmd/accounting-binance-trades-fetcher/main.go
```diff
@@ -0,0 +1,110 @@
...
+		return err
+	}
+
+	if c.Uint64(fromFlag) != 0 {
+		fromTime = timeutil.TimestampMsToTime(c.Uint64(fromFlag))
+	}
+
+	if c.Uint64(toFlag) != 0 {
+		toTime = timeutil.TimestampMsToTime(c.Uint64(toFlag))
+	}
```
favadi: should to time default to now?

lib/binance/binance_client.go
```diff
@@ -28,14 +27,14 @@ const (
 type Client struct {
 	APIKey      string
 	SecretKey   string
-	sugar       *zap.SugaredLogger
+	Sugar       *zap.SugaredLogger
```
favadi: why?

lib/binance/binance_client.go
```diff
@@ -47,28 +46,21 @@ func NewBinance(apiKey, secretKey string, sugar *zap.SugaredLogger, options ...O
 	clnt := &Client{
 		APIKey:    apiKey,
 		SecretKey: secretKey,
-		sugar:     sugar,
+		Sugar:     sugar,
```
favadi: why?

lib/binance/binance_client.go
```diff
@@ -178,13 +171,26 @@ func (bc *Client) GetTradeHistory(symbol string, fromID int64) ([]TradeHistory,
 	}
 
 	endpoint := fmt.Sprintf("%s/api/v3/myTrades", endpointPrefix)
+	params := map[string]string{
+		"symbol": symbol,
+	}
+
+	if fromID != -1 {
```
favadi: maybe using `fromID *int64` will be clearer. What if user give fromID=-2?

lib/binance/binance_client.go
```diff
@@ -167,7 +159,8 @@ func (bc *Client) sendRequest(method, endpoint string, params map[string]string,
 }
 
 //GetTradeHistory return history of trading on binance
-func (bc *Client) GetTradeHistory(symbol string, fromID int64) ([]TradeHistory, error) {
+//if fromID = -1 then function will not put fromId as a param
+func (bc *Client) GetTradeHistory(symbol string, fromID int64, fromTime, toTime time.Time) ([]TradeHistory, error) {
```
favadi: Do we really need both fromID and fromTime/toTime? The purpose of our fetcher is continually fetching everything from last stored until time.Now(). Maybe just fromID is enough? We could make the fetcher cli accept --from-id instead of from time / end time.

lib/binance/binance_stress_test.go
```diff
@@ -0,0 +1,43 @@
...
+)
+
+func TestBinance(t *testing.T) {
+	t.Skip("test disabled as require binance API key pair")
+	var (
+		apiKey    = os.Getenv("BINANCE_API_KEY")
+		secretKey = os.Getenv("BINANCE_SECRET_KEY")
+	)
+
+	logger, err := zap.NewDevelopment()
```
favadi: this could be replaced with testutil.MustNewDevelopmentLogger.

lib/binance/binance_stress_test.go
```diff
@@ -0,0 +1,43 @@
...
+
+	bn := NewBinance(apiKey, secretKey, sugar)
+
+	const requests = 10000
+	var g errgroup.Group
+	for i := 0; i < requests; i++ {
+		i := i
+		g.Go(func() error {
+			i := i
+			if _, err := bn.GetExchangeInfo(); err != nil {
```
favadi: we need to assert the response data?

lib/binance/flags.go
```diff
@@ -31,7 +30,7 @@ func NewCliFlags() []cli.Flag {
 			Name:   binanceRequestPerSecond,
 			Usage:  "binance request limit per second, default to 20 which etherscan's normal rate limit",
 			EnvVar: "BINANCE_REQUESTS_PER_SECOND",
-			Value:  20,
+			Value:  10,
```
favadi: why?

lib/binance/binance_stress_test.go
halink0803: No, stress test. I think we just need to check if client send request successfully or not and not break binance request limit. The assert response data we already did in binance_client_test.

lib/binance/flags.go
halink0803: 20 will break binance rate limit.

accounting/binance-fetcher/fetcher.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+func (f *Fetcher) getTradeHistoryWithRetry(symbol string, startTime, endTime time.Time) ([]binance.TradeHistory, error) {
+	var (
+		tradeHistoriesResponse []binance.TradeHistory
+		err                    error
+	)
+	for attempt := 0; attempt < f.attempt; attempt++ {
+		tradeHistoriesResponse, err = f.client.GetTradeHistory(symbol, -1, startTime, endTime)
+		if err == nil {
```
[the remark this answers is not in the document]
favadi: @halink0803 please remember to give an answer on the conversation after changes the codes.

accounting/binance-fetcher/fetcher.go
favadi: Ping?

lib/binance/flags.go
favadi: so the rate limiter is not working or what?

accounting/binance-fetcher/fetcher.go
halink0803: Actually, we already wait for retryDelay(default 2 min) when error return from Binance. If they still return error after 4 retry, we break the loop from batch requests. I think that all we can do because we cannot stop request already send in previous batch.

lib/binance/flags.go
halink0803: Yes, it seems the rate limit is not working. I tried with default batch size = 100.

accounting/binance-fetcher/fetcher.go
favadi: There is no point to stop already sent requests. It is fine then.

lib/binance/flags.go
favadi: I'll accept this but will recheck later.

PR #231, opened 2019-03-04
2540 reserve addresses

favadi:


accounting/reserve-addresses/http/addresses.go
```diff
@@ -0,0 +1,216 @@
...
+
+	"github.com/KyberNetwork/reserve-stats/accounting/common"
+	"github.com/KyberNetwork/reserve-stats/accounting/reserve-addresses/storage"
+	"github.com/KyberNetwork/reserve-stats/lib/httputil"
+	_ "github.com/KyberNetwork/reserve-stats/lib/httputil/validators" // import custom validator functions
+)
+
+type createInput struct {
+	Address     string `json:"address" binding:"required,isAddress"`
+	Type        string `json:"type"`
```
cyberliem: according to the doc:

> type: required, matches one of predefined types above

I suggest we change createInput struct binding to `required`

PR #234, opened 2019-03-05
2524_modified_rate_limiter

cyberliem:


lib/binance/binance_client.go
```diff
@@ -163,8 +163,11 @@ func (bc *Client) GetTradeHistory(symbol string, fromID int64) ([]TradeHistory,
 	)
 	const weight = 5
 	//Wait before creating the request to avoid timestamp request outside the recWindow
-	if err := bc.rateLimiter.WaitN(context.Background(), weight); err != nil {
-		return result, err
+	//Mimic  the leaky bucket algorithm to wait for 5  drop
+	for i := 0; i < weight; i++ {
```
favadi: we should do the same for other binance APIs method, despite that their cost = 1, so in the future if binance changes weight for API (which is already happened in the past), we can just change the weight const. Maybe make this loop in a common function.

cyberliem: fixed

PR #237, opened 2019-03-06
 refactor last block daily

favadi:


lib/lastblockdaily/last_block_daily.go
```diff
@@ -1,191 +1,253 @@
...
+		err   error
+	)
+
+	if lbr.lastResolved.Block == 0 {
+		// no last block of the day resolved, starts searching with:
+		// - from: first ever block timestamp
+		// - to: next day of configured start timestamp
+		start = firstBlock
+
+		if end, err = nextDayBlock(lbr.sugar, lbr.resolver, lbr.start, start); err != nil {
```
cyberliem: This logic is really bad for a big difference between nextDayBlock and earliest know block. The better approach should be binary search between earliest know block and latest known block for this. 
In fact,  I did a test:  on my branch, it tooks~ 7s for this with binary search. While on this branch, it took~ 22s to resolve. 

However, as this make the code much cleaner and the trade-off is one-off for first known block, we'll keep this implementation for now.

cyberliem: I created a redmine task for future reference https://redmine.knstats.com/issues/2686

PR #242, opened 2019-03-07
2536 huobi trade fetcher storage

cyberliem:


accounting/huobi/storage/postgres/postgres.go
```diff
@@ -0,0 +1,124 @@
...
+		tableNames: tableNames,
+	}, nil
+}
+
+//TearDown removes all the tables
+func (hdb *HuobiStorage) TearDown() error {
+	const dropFMT = `
+	DROP TABLE %s;
+	`
+	query := fmt.Sprintf(dropFMT, hdb.tableNames[huobiTradesTableName])
```
halink0803: This function remove all tables but the implementation remove only 1 table. You actually can loop through all tableNames and remove them.

cyberliem: if the tables are removed sequentially, we might need to cascade it. This is why currently I leave it as one drop query

accounting/cmd/accounting-huobi-trade-history-fetcher/main.go
```diff
@@ -7,6 +7,8 @@ import (
 	"time"
 
 	huobiFetcher "github.com/KyberNetwork/reserve-stats/accounting/huobi/fetcher"
+	postgres "github.com/KyberNetwork/reserve-stats/accounting/huobi/storage/postgres"
+
```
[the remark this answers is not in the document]
cyberliem: removed

accounting/cmd/accounting-huobi-trade-history-fetcher/main.go
```diff
@@ -6,19 +6,21 @@ import (
 	"os"
 	"time"
 
-	"github.com/urfave/cli"
-
 	huobiFetcher "github.com/KyberNetwork/reserve-stats/accounting/huobi/fetcher"
+	postgres "github.com/KyberNetwork/reserve-stats/accounting/huobi/storage/postgres"
```
favadi: redundant alias.

accounting/cmd/accounting-huobi-trade-history-fetcher/main.go
```diff
@@ -6,19 +6,21 @@ import (
...
 	libapp "github.com/KyberNetwork/reserve-stats/lib/app"
 	"github.com/KyberNetwork/reserve-stats/lib/huobi"
 	"github.com/KyberNetwork/reserve-stats/lib/timeutil"
+
+	"github.com/urfave/cli"
```
favadi: import group.

accounting/cmd/accounting-huobi-trade-history-fetcher/main.go
```diff
@@ -6,19 +6,21 @@ import (
...
 )
 
 const (
 	retryDelayFlag    = "retry-delay"
 	maxAttemptFlag    = "max-attempts"
 	defaultMaxAttempt = 3
 	defaultRetryDelay = time.Second
+	defaultPostGresDB = "reserve_stats"
```
favadi: should be lib/accounting/common.DefaultDB.

accounting/huobi/fetcher/fetcher.go
```diff
@@ -14,14 +15,20 @@ type Fetcher struct {
...
+func NewFetcher(sugar *zap.SugaredLogger, client huobi.Interface, retryDelay time.Duration, attempt int, strg ...storage.Interface) *Fetcher {
+	fetcher := &Fetcher{
 		sugar:      sugar,
 		client:     client,
 		retryDelay: retryDelay,
 		attempt:    attempt,
 	}
+	if len(strg) != 0 {
+		sugar.Info("fetcher is init with DB. Assigning...")
+		fetcher.db = strg[0]
```
favadi: we don't accept multiple storage.Interface, right? If that is the case, function option is better.

favadi: Actually, the fetcher should not be care about storage at all. It should have single responsibility is fetch and returns data.

accounting/huobi/fetcher/trade-history.go
```diff
@@ -129,3 +130,43 @@ func (fc *Fetcher) GetTradeHistory(from, to time.Time) (map[string][]huobi.Trade
 	})
 	return result, assertError
 }
+
+//GetAndStoreTradeHistory fetch all trade between from..to and store it in DB instead of output it from memory
+func (fc *Fetcher) GetAndStoreTradeHistory(from, to time.Time) error {
```
favadi: as commented before, when it is named Fetcher, it should only fetch the data. The data storing must be the responsibility of something else. Because we will need to reuse the fetcher somewhere else.

accounting/huobi/storage/postgres/postgres.go
```diff
@@ -0,0 +1,152 @@
...
+	tableNames map[string]string
+}
+
+// NewDB return the HuobiStorage instance. User must call Close() before exit.
+// tableNames is a list of string for tablename[huobitrades]. It can be optional
+func NewDB(sugar *zap.SugaredLogger, db *sqlx.DB, customTableNames ...string) (*HuobiStorage, error) {
+	const schemaFMT = `
+	CREATE TABLE IF NOT EXISTS %[1]s
+(
+	id bigint NOT NULL,
```
favadi: should its type be SERIAL? I don't see we explicitly created a sequence for it somewhere.

accounting/huobi/storage/postgres/postgres.go
```diff
@@ -0,0 +1,152 @@
...
+//HuobiStorage defines the object to store Huobi data
+type HuobiStorage struct {
+	sugar      *zap.SugaredLogger
+	db         *sqlx.DB
+	tableNames map[string]string
+}
+
+// NewDB return the HuobiStorage instance. User must call Close() before exit.
+// tableNames is a list of string for tablename[huobitrades]. It can be optional
+func NewDB(sugar *zap.SugaredLogger, db *sqlx.DB, customTableNames ...string) (*HuobiStorage, error) {
```
favadi: `customTableNames ...string`, functional option is better. It is now user friendly to allow passing multiple custom table names but returns an error for it.

accounting/huobi/storage/postgres/postgres.go
```diff
@@ -0,0 +1,152 @@
...
+	const schemaFMT = `
+	CREATE TABLE IF NOT EXISTS %[1]s
+(
+	id bigint NOT NULL,
+	time TIMESTAMP NOT NULL,
```
favadi: no index for timestamp?

accounting/huobi/storage/postgres/postgres.go
```diff
@@ -0,0 +1,152 @@
...
+	symbol TEXT,
```
favadi: could you double check if we need to search for symbol anywhere? It is already in the JSON field?

accounting/huobi/storage/postgres/postgres.go
favadi: awn, I checked the field of json data and it already has both timestamp and symbol. We should remove this and create an index for json field timestamp to avoid data duplication.

accounting/huobi/storage/postgres/postgres.go
```diff
@@ -0,0 +1,152 @@
...
+	query := fmt.Sprintf(updateStmt,
+		hdb.tableNames[huobiTradesTableName],
+	)
+	data, err := json.Marshal(trade)
+	if err != nil {
+		return err
+	}
+	if err != nil {
+		return err
+	}
```
favadi: the error is handled twice.

accounting/cmd/accounting-huobi-trade-history-fetcher/main.go
cyberliem: I separated internal packages and external packages already?

accounting/cmd/accounting-huobi-trade-history-fetcher/main.go
cyberliem: removed

accounting/cmd/accounting-huobi-trade-history-fetcher/main.go
cyberliem: changed.

accounting/huobi/storage/postgres/postgres.go
cyberliem: This ID is passed down from Huobi, we don't do auto increment for it. BIGSERIAL is defined as "64-bit auto-increment integer data" so I use BIGINT instead.

accounting/huobi/fetcher/fetcher.go
cyberliem: removed storage from fetcher

accounting/huobi/fetcher/trade-history.go
cyberliem: removed

accounting/huobi/storage/postgres/postgres.go
cyberliem: change to options

accounting/huobi/storage/postgres/postgres.go
cyberliem: I created index for timestamp as:
```sql CREATE INDEX IF NOT EXISTS time_idx ON huobi_trades ((data ->> 'created-at')); 
```

As is, postGres use 5 types of INDEX: B-tree(default), Hash, Gist, SP-Gist, Gin. I choose the default B-tree since we only need to index a specific time field. Other Indexes are ruled out:
- Hash Index is only good for equal comparison query. What's we doing here is range query. 
- Using Gin might be faster if we need to fully index the json object for static query- which is what we need ( these data will be look-uped more frequently than updated). However, we only need to index a specific time field, GIN is wasteful as it index the whole JSON objects and its fields. 
- Using GIST, which is similar to GIN- index for full text search as their [doc](https://www.postgresql.org/docs/9.3/textsearch-indexes.html) specified. GIST is for dynamic data update (it update faster than GIN but lookup slower). 
-  SP-GiST is like GiST, but optimized for operator classes for two-dimensional points.
- BRIN on theory suits our need better, that is, data in "sequential order". However, real test show that although on first run, BRIN is faster than B-tree. Subsequent run B-tree is faster. This might due to the caching policy of Postgres: the result is shown as follow:
```
B-tree first run:
"Bitmap Heap Scan on huobi_trades  (cost=82.18..689.29 rows=3341 width=421) (actual time=21.031..22.746 rows=7701 loops=1)"
"  Recheck Cond: ((data ->> 'created-at'::text) >= '1552549990000'::text)"
"  Heap Blocks: exact=428"
"  ->  Bitmap Index Scan on huobi_trades_time_idx3  (cost=0.00..81.34 rows=3341 width=0) (actual time=20.969..20.969 rows=7701 loops=1)"
"        Index Cond: ((data ->> 'created-at'::text) >= '1552549990000'::text)"
"Planning time: 9.181 ms"
"Execution time: 30.019 ms"

B-tree subsequence run:
"Bitmap Heap Scan on huobi_trades  (cost=82.18..689.29 rows=3341 width=421) (actual time=21.031..22.746 rows=7701 loops=1)"
"  Recheck Cond: ((data ->> 'created-at'::text) >= '1552549990000'::text)"
"  Heap Blocks: exact=428"
"  ->  Bitmap Index Scan on huobi_trades_time_idx3  (cost=0.00..81.34 rows=3341 width=0) (actual time=20.969..20.969 rows=7701 loops=1)"
"        Index Cond: ((data ->> 'created-at'::text) >= '1552549990000'::text)"
"Planning time: 9.181 ms"
"Execution time: 30.019 ms"

BRIN tree run (similar on every run):
"Bitmap Heap Scan on huobi_trades  (cost=37.89..645.01 rows=3341 width=421) (actual time=0.604..6.078 rows=7701 loops=1)"
"  Recheck Cond: ((data ->> 'created-at'::text) >= '1552549990000'::text)"
"  Rows Removed by Index Recheck: 21"
"  Heap Blocks: lossy=429"
"  ->  Bitmap Index Scan on huobi_trades_time_idx3  (cost=0.00..37.05 rows=3341 width=0) (actual time=0.573..0.573 rows=5120 loops=1)"
"        Index Cond: ((data ->> 'created-at'::text) >= '1552549990000'::text)"
"Planning time: 1.399 ms"
"Execution time: 7.501 ms"
```

accounting/huobi/storage/postgres/postgres.go
cyberliem: I removed the field

accounting/huobi/storage/postgres/postgres.go
cyberliem: Removed 1 handler

accounting/huobi/storage/postgres/postgres.go
favadi: BTree should be fine most of the case, so it is OK to go with it for now.

Just curious, where is the source for?

> However, we only need to index a specific time field, GIN is wasteful as it index the whole JSON objects and its fields.

cyberliem: > BTree should be fine most of the case, so it is OK to go with it for now.
> 
> Just curious, where is the source for?
> 

In their doc name https://www.postgresql.org/docs/9.3/textsearch-indexes.html, the mentioned at the first row that There are two kinds of indexes that can be used to speed up full text searches.

In addition to that, on create GIN index it specified:
Creates a GIN (Generalized Inverted Index)-based index. The column **must** be of tsvector type.

Further reading ong tsvector type https://www.postgresql.org/docs/10/datatype-textsearch.html
it specified that this is some kind similar to a structure to hold mutiple strings, not a single strings.

accounting/cmd/accounting-huobi-trade-history-fetcher/main.go
favadi: it is blank line in L14. For import group, we only need to separate builtin packages and external packages.

```go
import (
    fmt

    github.com/urfave/cli
    github.com/KyberNetwork/reserve-stats/lib/huobi
)
```

Even better if we can separate builtin, external, internal packages like.

```go
import (
    fmt

    github.com/urfave/cli

    github.com/KyberNetwork/reserve-stats/lib/huobi
)
```

accounting/huobi/storage/postgres/postgres.go
favadi: > This ID is passed down from Huobi, we don't do auto increment for it. BIGSERIAL is defined as "64-bit auto-increment integer data" so I use BIGINT instead.

this is correct, sorry for the fud.

accounting/huobi/storage/postgres/postgres.go
favadi: > However, we only need to index a specific time field, GIN is wasteful as it index the whole JSON objects and its fields.

All above quotes does not prove that GIN always index whole JSON object. ANW, it is not important at this moment, we'll go with btree.

PR #243, opened 2019-03-08
2534 accounting-binance-trades-fetcher-review

favadi:


lib/binance/rate_limiter.go
```diff
@@ -0,0 +1,65 @@
...
+	wafLimiter := rate.NewLimiter(rate.Limit(wafLimit), 1)
+	return &RateLimiter{
+		m:           sync.Mutex{},
+		hardLimiter: hardLimiter,
+		wafLimiter:  wafLimiter,
+	}
+}
+
+// WaitN waits until enough resources are available for a request with given weight.
+func (r *RateLimiter) WaitN(ctx context.Context, n int) error {
```
cyberliem: The approach in this function might slow down the services. Our intent is to make a request wait for a period that satisfy both of the Limiter. This sequentially approach make a request wait for 2 periods, each for a limiter.

PR #244, opened 2019-03-08
2538 huobi withdrawal service

cyberliem:


accounting/huobi/fetcher/withdrawal-history.go
```diff
@@ -0,0 +1,104 @@
...
+		errGroup    errgroup.Group
+	)
+
+	symbols, err := fc.client.GetCurrencies()
+	if err != nil {
+		return result, err
+	}
+	for _, sym := range symbols {
+		errGroup.Go(
+			func(symbol string) func() error {
```
favadi: the function inside should be moved to a separate function.

accounting/huobi/fetcher/withdrawal-history.go
```diff
@@ -0,0 +1,104 @@
...
+)
+
+func retryGetWithdrawal(fn func(string, uint64) (huobi.WithdrawHistoryList, error), symbol string, fromID uint64, attempt int, retryDelay time.Duration) (huobi.WithdrawHistoryList, error) {
+	var (
+		result huobi.WithdrawHistoryList
+		err    error
+	)
+	for i := 0; i < attempt; i++ {
+		result, err = fn(symbol, fromID)
+		if err == nil {
```
favadi: the error should be logged if ! nil.

accounting/huobi/fetcher/withdrawal-history.go
```diff
@@ -0,0 +1,104 @@
...
+		fetchResult = sync.Map{}
+		assertError error
+		errGroup    errgroup.Group
+	)
+
+	symbols, err := fc.client.GetCurrencies()
+	if err != nil {
+		return result, err
+	}
+	for _, sym := range symbols {
```
favadi: we don't need to do it in batch and handle 429?

lib/huobi/huobi_client.go
```diff
@@ -248,3 +248,28 @@ func (hc *Client) GetSymbolsPair() ([]Symbol, error) {
...
+		false,
+	)
+	if err != nil {
+		return reply.Data, err
+	}
+	err = json.Unmarshal(res, &reply)
+	if err != nil {
+		return reply.Data, err
+	}
+	if reply.Status != "ok" {
```
favadi: please check is we check status "ok" everywhere already? If yes, "ok" should be a const.

accounting/huobi/fetcher/withdrawal-history.go
```diff
@@ -0,0 +1,104 @@
...
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/KyberNetwork/reserve-stats/lib/huobi"
+
+	"golang.org/x/sync/errgroup"
+)
+
+func retryGetWithdrawal(fn func(string, uint64) (huobi.WithdrawHistoryList, error), symbol string, fromID uint64, attempt int, retryDelay time.Duration) (huobi.WithdrawHistoryList, error) {
```
favadi: please make the function inside a type.

accounting/huobi/fetcher/withdrawal-history.go
```diff
@@ -0,0 +1,104 @@
...
+	return result, err
+}
+
+func (fc *Fetcher) getWithdrawHistoryWithSymbol(symbol string, fromID uint64) ([]huobi.WithdrawHistory, error) {
+	var (
+		nextFromID = fromID
+		result     []huobi.WithdrawHistory
+	)
+	for {
+		tradeHistoriesResponse, err := retryGetWithdrawal(fc.client.GetWithdrawHistory, symbol, nextFromID, fc.attempt, fc.retryDelay)
```
favadi: it will be clearer if we have a `getWithdrawHistoryWithSymbol` function and feed it to `retryGetWithdrawal`.

accounting/huobi/fetcher/withdrawal-history.go
```diff
@@ -0,0 +1,104 @@
...
+		// while result != empty, get trades latest time to toTime
+		if len(tradeHistoriesResponse.Data) == 0 {
+			break
+		}
+		result = append(result, tradeHistoriesResponse.Data...)
+		lastWithdrawal := tradeHistoriesResponse.Data[0]
+		nextFromID = lastWithdrawal.ID + 1
+	}
+
+	return result, nil
```
favadi: anw, could you try dropping the for loop and use recursive to see if the codes looks shorter?

accounting/huobi/fetcher/withdrawal-history.go
cyberliem: the code is much more readable this way because:

This saves us from a lot of params passing.
It show clear proceeding of how the data/err is handle inside.

accounting/huobi/fetcher/withdrawal-history.go
cyberliem: logged.

accounting/huobi/fetcher/withdrawal-history.go
cyberliem: as testing show, huobi doesn't send 429 and doesn't block request if we exceed the rate limit

lib/huobi/huobi_client.go
cyberliem: changed

accounting/huobi/fetcher/withdrawal-history.go
cyberliem: made

accounting/huobi/fetcher/withdrawal-history.go
cyberliem: the retry logic only handle individual call (mostly to stop huobi from rarely happens random excess limit). getWithdrawHistoryWithSymbol otherwise handles the logic of continuous fetching till last withdrawal

accounting/huobi/fetcher/withdrawal-history.go
cyberliem: this function is meant to do continuously fetching, not "divide and conquer" fetching, using recursive here only increase its stack side grow and more complicated in mem handling.

accounting/huobi/fetcher/withdrawal-history.go
```diff
@@ -0,0 +1,109 @@
...
+		result huobi.WithdrawHistoryList
+		err    error
+		logger = fc.sugar.With("func", "accounting/huobi/fetcher/withdrawal-history.retryGetWithdrawal")
+	)
+	for i := 0; i < fc.attempt; i++ {
+		result, err = fn(symbol, fromID)
+		if err == nil {
+			return result, nil
+		}
+		logger.Debugw("fail to fetch withdrawal history", "error", err, "attempt", i+1)
```
favadi: I'll change the log level to warn myself.

accounting/huobi/fetcher/withdrawal-history.go
favadi: > This saves us from a lot of params passing.

I don't see how params passing is a bad thing.

>  It show clear proceeding of how the data/err is handle inside.

We can still see the this wherever the codes place. 

Anw, the codes look simple enough, so just keep it.

accounting/huobi/fetcher/withdrawal-history.go
favadi: > this function is meant to do continuously fetching, not "divide and conquer" fetching,

It is not very clear how they are different.

Anw, it is not important.

accounting/huobi/fetcher/withdrawal-history.go
favadi: did that in 39b6328378d9e04b26133eb269aee15fe0034ee1.

PR #246, opened 2019-03-11
2707 Update huobi response and log in test

halink0803:


lib/huobi/huobi_client_test.go
```diff
@@ -7,15 +7,15 @@ import (
...
+	// "github.com/KyberNetwork/reserve-stats/lib/testutil"
 
 	"github.com/stretchr/testify/assert"
 	"go.uber.org/zap"
 	"golang.org/x/time/rate"
 )
 
 func TestHuobiClient(t *testing.T) {
-	testutil.SkipExternal(t)
+	// testutil.SkipExternal(t)
```
favadi: just always skip this test.

lib/huobi/types.go
```diff
@@ -41,24 +41,17 @@ type TradeHistoryList struct {
...
-	TxHash            string  `json:"tx-hash"`
-	Chain             string  `json:"chain"`
-	Extra             string  `json:"extra"`
+	ID        uint64  `json:"id"`
+	Type      string  `json:"type"`
+	Currency  string  `json:"currency"`
+	TxHash    string  `json:"tx-hash"`
+	Amount    float64 `json:"amount"`
+	Address   string  `json:"address"`
+	AddrTag   string  `json:"address-tag"`
```
favadi: please rename it to AddressTag to match json field, we use `Address` to above field already.

PR #248, opened 2019-03-12
add pgadmin4 to docker-compose

ductm54:


docker-compose.yml
```diff
@@ -23,4 +23,12 @@ services:
     volumes:
       - './data/redis:/var/lib/redis/'
     ports:
-      - '127.0.0.1:6379:6379' 
\ No newline at end of file
+      - '127.0.0.1:6379:6379'
+  pgadmin4:
+    image: dpage/pgadmin4
+    ports:
+        - '127.0.0.1:60978:60978'
```
favadi: From https://hub.docker.com/r/dpage/pgadmin4, I think the default listening port is 80.

docker-compose.yml
```diff
@@ -23,4 +23,12 @@ services:
...
+    environment:
+        PGADMIN_LISTEN_PORT: 60978
+        PGADMIN_DEFAULT_EMAIL: postgres@localhost
+        PGADMIN_DEFAULT_PASSWORD: secret
```
favadi: could you make email, password: reserve_stats@localhost/reserve_stats just like default password?

ductm54: changed

docker-compose.yml
ductm54: updated

PR #250, opened 2019-03-13
2531-accounting-reserve-tx-fetcher

favadi:


accounting/cmd/accounting-reserve-transaction-fetcher/main.go
```diff
@@ -0,0 +1,114 @@
...
+
+	logger, err := libapp.NewLogger(c)
+	if err != nil {
+		return err
+	}
+	defer logger.Sync()
+
+	sugar := logger.Sugar()
+
+	addrs := c.StringSlice(addressesFlag)
```
cyberliem: Maybe adding somekind of verification for addrs. I ran with empty addrs flags and nothing is returned. Ran with a non erc-20 address it doesn't raise any error.

accounting/reserve-transaction-fetcher/fetcher/fetcher.go
```diff
@@ -0,0 +1,210 @@
...
+			return nil, err
+		}
+		results := make([]interface{}, len(internalTxs))
+		for i, v := range internalTxs {
+			results[i] = v
+		}
+		return results, nil
+	})
+
+	results, err := f.fetch(fn, addr, from, to)
```
cyberliem: Is the conversion from *big.Int to *int safe?

accounting/reserve-transaction-fetcher/fetcher/fetcher.go
```diff
@@ -0,0 +1,210 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	txs := make([]common.InternalTx, len(results))
+	for i, v := range results {
+		tx := v.(etherscan.InternalTx)
```
cyberliem: should do tx, ok := v.. for type assertion and make sure ok is not false.

accounting/reserve-transaction-fetcher/fetcher/fetcher.go
```diff
@@ -0,0 +1,210 @@
...
+		Value:           tx.Value.Int(),
+		Gas:             tx.Gas,
+		GasUsed:         tx.GasUsed,
+		GasPrice:        tx.GasPrice.Int(),
+	}
+}
+
+// ERC20Transfer returns all ERC20 transfers of given address between given block range.
+func (f *EtherscanTransactionFetcher) ERC20Transfer(addr ethereum.Address, from, to *big.Int) ([]common.ERC20Transfer, error) {
+	fn := newFetchFunction("transfer", func(address string, startBlock *int, endBlock *int, page int, offset int) ([]interface{}, error) {
```
cyberliem: A really nice approach to group these fetching functions. However, as the number of tx are huges, this approach cause much more overhead on memory allocation (double allocation: one for []interface and one for []results). If your stress test yesterday proves this is no problem, please ignore this comment.

accounting/cmd/accounting-reserve-transaction-fetcher/main.go
favadi: added that in 62953b4db444485dbdc7c4003f039f904b7347fe.

accounting/reserve-transaction-fetcher/fetcher/fetcher.go
favadi: no, but etherscan only supports int. I added a safe guard in 33fc9ee37134ebb507cae1f77f9c664d13ee6bec.

accounting/reserve-transaction-fetcher/fetcher/fetcher.go
favadi: In this case, just panic is better because it can only happens if we messed up the InternalTx --> interface{} --> InternalTx. That means, programmer error.

PR #251, opened 2019-03-14
improve accounting rates postgres tests

favadi:


accounting/reserve-rate/fetcher/fetcher.go
```diff
@@ -185,20 +185,29 @@ func retryFetchTokenRate(maxAttempt int,
...
+		result = make(map[string]map[string]float64)
 		err    error
 		logger = sugar.With("function", "accounting/reserve-rate/fetcher/fetcher.retryFetchTokenRate", "block", block)
 	)
 
 	for i := 0; i < maxAttempt; i++ {
-		result, err = rsvRateCrawler.GetReserveRates(block)
-		if err == nil {
-			return result, nil
+		// TODO: GetReserveRates might fetch too more data than we need here?
```
cyberliem: This requires call to Wrapper contract's  GetReserveRates and it returns as whole. Diminishing the extra datas doesn't improve much of the performance or code quality

accounting/reserve-rate/fetcher/fetcher.go
```diff
@@ -185,20 +185,29 @@ func retryFetchTokenRate(maxAttempt int,
...
+			logger.Debugw("failed to fetch reserve rate", "attempt", i, "error", err)
+			time.Sleep(retryInterval)
+			continue
 		}
-		logger.Debugw("failed to fetch reserve rate", "attempt", i, "error", err)
-		time.Sleep(retryInterval)
+		for k1 := range rates {
+			result[k1] = make(map[string]float64)
+			for k2 := range rates[k1] {
+				// TODO: is it really buy reserve rate?
```
cyberliem: Regarding an example rate from blockchain for ETH-KNC:
  "ETH-KNC": {
    "buy_reserve_rate": 904.948930826676,
    "buy_sanity_rate": 0,
    "sell_reserve_rate": 0.001076798015,
    "sell_sanity_rate": 0
  },

Buy reserve rate mean to buy an ETH(our quotes), we need ~905 KNC. Regarding our spec:

> 
KNC/ETH rate means how much KNC I have to sell to get 1 ETH. So if KNC/ETH bid is 0.0166667 on Kyber,
KC/ETH rate is 600

accounting/reserve-rate/storage/interface.go
```diff
@@ -1,16 +1,18 @@
...
 	"time"
+
+	"github.com/KyberNetwork/reserve-stats/lib/lastblockdaily"
 )
 
 //Interface defines a set of interface for reserve rate storage, which can be implemented by any DB
 type Interface interface {
-	UpdateRatesRecords(lastblockdaily.BlockInfo, map[string]map[string]common.ReserveRateEntry) error
-	UpdateETHUSDPrice(blockInfo lastblockdaily.BlockInfo, ethusdRate float64) error
+	// TODO: should we merge UpdateETHUSDPrice and UpdateETHUSDPrice?
```
cyberliem: ETHUSDPrice does not come with reserve address. This meant if we merge it, we're going to duplicate ETH-USD price across all reserves.

The Upside of merging will atomic all rate operation and reduce amount of codes in the DB ops. I'm not sure if duplication will pay off for this.

accounting/reserve-rate/storage/postgres/accounting_rates_postgres.go
```diff
@@ -62,6 +62,9 @@ func (rdb *RatesStorage) Close() error {
 }
 
 //getTokenBase from pair return token and base from base-token pair
+// TODO: parsing the pair string from storage doesn't look right,
+//  as all the database implementations will need to redo it.
+//  Maybe the fetcher should do it before passing to storage?
```
cyberliem: As the signature of all getReserveRate function return an object as
```golang
map[reserverAddress]map[Pair]Rate
```
Changing this requires changing a lot. We can do this later.

accounting/reserve-rate/fetcher/fetcher.go
favadi: At least it does save bandwidth, but we could do the improvement later, please create a task and remove TODO.

accounting/reserve-rate/storage/interface.go
favadi: > ETHUSDPrice does not come with reserve address. This meant if we merge it, we're going to duplicate ETH-USD price across all reserves.

I don't see why we need to duplicate, just make the method signature like.

```go
Update(blockInfo lastblockdaily.BlockInfo, ethusdRate float64, map[string]map[string]common.ReserveRateEntry)
```

and we can still keep how the rate is stored.

accounting/reserve-rate/storage/postgres/accounting_rates_postgres.go
favadi: Please create a task and remove the TODO.

accounting/reserve-rate/storage/postgres/accounting_rates_postgres.go
```diff
@@ -313,47 +321,32 @@ func (rdb *RatesStorage) UpdateETHUSDPrice(blockInfo lastblockdaily.BlockInfo, e
...
-	if err != nil {
-		return err
-	}
 
 	return err
 }
 
 //GetLastResolvedBlockInfo return block info of the rate with latest timestamp
-func (rdb *RatesStorage) GetLastResolvedBlockInfo() (lastblockdaily.BlockInfo, error) {
+func (rdb *RatesStorage) GetLastResolvedBlockInfo(reserveAddr ethereum.Address) (lastblockdaily.BlockInfo, error) {
```
cyberliem: This function is now accept reserveAddress as extra param and it return the latest block info with that reserve Address only. To do this, these change are made:
-- UpdateRates is now update ETHUSD rate as well in 1 tx. 
-- Query the lastBlockInfo only by token_rates table, not from usd_rates anymore

PR #256, opened 2019-03-19
2546 Accounting binance withdraw storage

halink0803:


accounting/binance-storage/binance_storage.go
```diff
@@ -0,0 +1,190 @@
...
+)
+
+//BinanceStorage is storage for binance fetcher including trade history and withdraw history
+type BinanceStorage struct {
+	sugar *zap.SugaredLogger
+	db    *sqlx.DB
+}
+
+//NewDB return a new instance of binance storage
+func NewDB(sugar *zap.SugaredLogger, db *sqlx.DB, tableName, timeIndexField, idType string) (*BinanceStorage, error) {
```
cyberliem: better to have option WithTradeTableName, WithWithdrawTableName to better define the db. 

In addition to that, why is timeIndexField and idType is required as input? as [line 154](https://github.com/KyberNetwork/reserve-stats/pull/256/files#diff-b45d42bd4862a0eca43b43e5d2d99640R150) suggest, the fields name are fixed as they are not input in to the fmt.Sprintf.

accounting/binance-storage/binance_storage.go
```diff
@@ -0,0 +1,190 @@
...
+func (bd *BinanceStorage) DeleteTable(tableName string) error {
+	query := fmt.Sprintf("DROP TABLE %s", tableName)
+	if _, err := bd.db.Exec(query); err != nil {
+		return err
+	}
+	return nil
+}
+
+//UpdateTradeHistory save trade history into a postgres db
+func (bd *BinanceStorage) UpdateTradeHistory(trades []binance.TradeHistory) error {
```
cyberliem: Why is there updateTradeHistory here? Is this branch supposed to be merged into accounting-trade-history-storage or develop?

accounting/binance-storage/binance_storage.go
```diff
@@ -0,0 +1,190 @@
...
+
+	tx, err := bd.db.Beginx()
+	if err != nil {
+		return err
+	}
+
+	defer pgsql.CommitOrRollback(tx, bd.sugar, &err)
+
+	for _, withdraw := range withdrawHistories {
+		query := fmt.Sprintf(updateQuery, binanceWithdrawTable)
```
cyberliem: should take query out of the loop and log it

accounting/binance-storage/interface.go
```diff
@@ -1,5 +1,15 @@
...
+	"time"
+
+	"github.com/KyberNetwork/reserve-stats/lib/binance"
+)
+
 //Interface is inteface for binance storage
 type Interface interface {
+	UpdateTradeHistory([]binance.TradeHistory) error
+	GetTradeHistory(fromTime, toTime time.Time) ([]binance.TradeHistory, error)
+	UpdateWithdrawHistory(map[string]binance.WithdrawHistory) error
```
cyberliem: As specification required, the cex_trades will only required binaceStorage.UpdateTradeHistory and binaceStorage.GetTradeHistory as well as Huobi's similar interface. In my opinion, grouping Binance's storage methods on both trades and withdrawal is not conform to current specification.

I propose a different approach: create two storage packages for trades and withdrawal, as demonstrate in following folder structure:
```
binance
-->fetcher
-->storage
     -->trades
     -->withdrawal
```

Please advise @favadi @halink0803

accounting/binance-storage/binance_storage.go
halink0803: Yes, it is suppose to merge into accounting-trade-history-storage branch. Let me change the base.

accounting/binance-storage/interface.go
halink0803: Yes, I think it is easy to separate the interface.

accounting/binance-storage/binance_storage.go
halink0803: I used to use NewDB for both init TradeHistoryStorage and WithdrawStorage and idFields and timeIndexField are different from them. That why it is needed to set as variable. Changed now.

accounting/binance-storage/trade-storage/trade_storage.go
```diff
@@ -130,3 +130,33 @@ func (bd *BinanceStorage) GetTradeHistory(fromTime, toTime time.Time) ([]binance
 
 	return result, nil
 }
+
+//UpdateWithdrawHistory save withdraw history to db
+func (bd *BinanceStorage) UpdateWithdrawHistory(withdrawHistories map[string]binance.WithdrawHistory) error {
```
cyberliem: Why there is UpdateWithdrawHistory in trade_storage?

halink0803: Removed

accounting/cmd/accounting-binance-withdraw-fetcher/main.go
```diff
@@ -109,10 +109,26 @@ func run(c *cli.Context) error {
...
+	}(&err)
+
 	withdrawHistory, err := binanceFetcher.GetWithdrawHistory(fromTime, toTime)
 	if err != nil {
 		return err
 	}
 
-	return binanceStorage.UpdateWithdrawHistory(withdrawHistory)
+	err = binanceStorage.UpdateWithdrawHistory(withdrawHistory)
+	if err != nil {
```
cyberliem: just return err is enough it would seem, whatever err is, the defer func will handle it

halink0803: changed.

accounting/cmd/accounting-binance-withdraw-fetcher/main.go
```diff
@@ -95,12 +99,26 @@ func run(c *cli.Context) error {
...
 	if err != nil {
 		return err
 	}
 
-	sugar.Infow("withdraw history", "value", withdrawHistory)
+	defer func() {
+		if cErr := binanceStorage.Close(); cErr != nil {
+			sugar.Errorf("Close database error", "error", cErr)
+		}
+	}()
```
favadi: we should also close it before return.

```go
	if err = binanceStorage.UpdateTradeHistory(tradeHistories); err != nil {
		return err
	}
	return binanceStorage.Close()
```

accounting/binance-storage/withdraw-storage/withdraw_storage.go
```diff
@@ -0,0 +1,134 @@
...
+func (bd *BinanceStorage) UpdateWithdrawHistory(withdrawHistories map[string]binance.WithdrawHistory) (err error) {
+	var (
+		logger       = bd.sugar.With("func", "accounting/binance_storage.UpdateWithdrawHistory")
+		withdrawJSON []byte
+	)
+	const updateQuery = `INSERT INTO %[1]s (id, data)
+	VALUES(
+		$1,
+		$2
+	) ON CONFLICT ON CONSTRAINT %[1]s_pk DO NOTHING;
```
favadi: we probably could do multiple values insert here. I'll create task to do later.

accounting/binance-storage/withdraw-storage/withdraw_storage_test.go
```diff
@@ -0,0 +1,109 @@
...
+		postgresPort,
+		postgresUser,
+		postgresPassword,
+		postgresDatabase,
+	)
+	db, err := sqlx.Connect("postgres", connStr)
+	if err != nil {
+		return nil, err
+	}
+	return NewDB(sugar, db, tableName)
```
favadi: please use `testutil.MustNewDevelopmentDB`.

accounting/binance-storage/withdraw-storage/interface.go
```diff
@@ -0,0 +1,13 @@
...
+
+import (
+	"time"
+
+	"github.com/KyberNetwork/reserve-stats/lib/binance"
+)
+
+//Interface is inteface for binance storage
+type Interface interface {
+	UpdateWithdrawHistory(map[string]binance.WithdrawHistory) error
```
favadi: not sure what is the key of this map? I think it should just be []binance.WithdrawHistory.

accounting/cmd/accounting-binance-withdraw-fetcher/main.go
```diff
@@ -51,6 +54,7 @@ func main() {
 
 	app.Flags = append(app.Flags, binance.NewCliFlags()...)
 	app.Flags = append(app.Flags, timeutil.NewMilliTimeRangeCliFlags()...)
```
favadi: I think we missing default values for from time to time somewhere. Please make default from time 2018-01-01 (should be a const somewhere already) and default to time now().

accounting/binance-storage/withdraw-storage/interface.go
halink0803: We need to avoid duplicate record because we fetcher fromTime and toTime. However, if we handle duplicate record when insert into database then it is ok to use []binance.WithdrawHistory.

accounting/cmd/accounting-binance-withdraw-fetcher/main.go
halink0803: Because if we don't input fromTime then binance will return from latest withdraw, then you said it is not necessary to set default for fromTime. But it makes no harm if we set default fromTime to 2018-01-01.

accounting/cmd/accounting-binance-withdraw-fetcher/main.go
halink0803: Added

accounting/binance-storage/withdraw-storage/withdraw_storage_test.go
halink0803: Changed

accounting/binance-storage/withdraw-storage/interface.go
favadi: the database insertion need to be able to handle duplicated insertion anw.

accounting/cmd/accounting-binance-withdraw-fetcher/main.go
favadi: I tried with emtpy from/to flag and the program just exit with invalid flag error
```
2019/03/25 11:55:19 empty flag
```

I don't even know which flag is empty.

halink0803: Yes, it comes from the lib. I change the lib to make it optional.

accounting/binance-storage/withdraw-storage/interface.go
halink0803: Updated.

accounting/binance-fetcher/fetcher.go
```diff
@@ -149,30 +147,34 @@ func (f *Fetcher) getWithdrawHistoryWithRetry(startTime, endTime time.Time) (bin
 	return withdrawHistory, err
 }
 
-func appendResult(result map[string]binance.WithdrawHistory, withdrawList []binance.WithdrawHistory) time.Time {
+func appendResult(result []binance.WithdrawHistory, withdrawList []binance.WithdrawHistory, lastID string) ([]binance.WithdrawHistory, time.Time, string) {
```
favadi: as API docs: https://github.com/binance-exchange/binance-official-api-docs/blob/master/wapi-api.md#withdraw-history-user_data, I don't see limit, paging anywhere. I tested with my API key and it successfully returns data for >8000 trades. Probably safe to remove this.

halink0803: Removed it

PR #263, opened 2019-03-25
2530 wallet erc20 txs

cyberliem:


accounting/cmd/accounting-wallet-erc20/main.go
```diff
@@ -0,0 +1,110 @@
...
+	app.Flags = append(app.Flags,
+		cli.StringSliceFlag{
+			Name:   walletAddressesFlag,
+			EnvVar: "WALLET_ADDRESSES",
+			Usage:  "list of wallet addresses to fetch transactions",
+		},
+		cli.StringFlag{
+			Name:   fromBlockFlag,
+			Usage:  "Fetch transactions from block",
+			EnvVar: "FROM_BLOCK",
```
halink0803: In etherscan api, from block is not required. Why it is required here? Should it have a default value?

accounting/cmd/accounting-wallet-erc20/main.go
```diff
@@ -0,0 +1,110 @@
...
+		},
+		cli.StringFlag{
+			Name:   toBlockFlag,
+			Usage:  "Fetch transactions to block",
+			EnvVar: "TO_BLOCK",
```
halink0803: In etherscan api, to block is not required. Why it is required here? Should it have a default value?

cyberliem: etherscan API ERC20-transfer has to block param. No specification on default value for to block. To block can get till latest block, but I don't see any use case here since without DB implementation, daemon mode is not effective?

halink0803: Yes, I know it has toBlock param but it is not required. I think it should be optional here as well. It could be left blank, you can test as it could response till latest block.

accounting/cmd/accounting-wallet-erc20/main.go
cyberliem: you're sure about this? etherscan API ERC20-transfer has from block param as can be seen here https://etherscan.io/apis#accounts. 

What do you suggest to be default value for it when DB is not implemented yet?

favadi: we could agree on a sane from_block later. For now it is OK.

accounting/cmd/accounting-wallet-erc20/main.go
favadi: we do have to block default to latest known block everywhere, so it should be nice to have it here.

accounting/wallet-fetcher/fetcher.go
```diff
@@ -0,0 +1,83 @@
+package fetcher
```
favadi: we should name this package wallet-erc20-transaction-fetcher or something, or it might became confusing later.

accounting/wallet-erc20/storage/postgres/postgres.go
```diff
@@ -0,0 +1,192 @@
...
+	const schemaFMT = `
+	CREATE TABLE IF NOT EXISTS %[1]s
+(
+	id BIGSERIAL NOT NULL,
+	data JSONB NOT NULL UNIQUE,
+	CONSTRAINT %[1]s_pk PRIMARY KEY(id)
+) ;
+
+CREATE INDEX IF NOT EXISTS %[1]s_time_idx ON %[1]s 
+((data ->> 'timeStamp'),(data ->> 'contractAddress'),(data ->> 'from'),(data ->> 'from'));
```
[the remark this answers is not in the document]
favadi: fixed in f7dc252ab97f0b9f1d6e3d3133898d764cf25f74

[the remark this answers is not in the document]
cyberliem: confirm fixed

PR #266, opened 2019-03-26
2552 Accounting gateway

halink0803:


accounting/cmd/accounting-gateway/main.go
```diff
@@ -0,0 +1,128 @@
...
+			Name:   cexTradeAPIURLFlag,
+			Usage:  "cex trade api url",
+			Value:  defaultCexTradeAPIValue,
+			EnvVar: "CEX_TRADE_URL",
+		},
+		cli.StringFlag{
+			Name:   reserveAddressesAPIURLFlag,
+			Usage:  "reserve addresses api url",
+			Value:  defaultReserveAddressAPIValue,
+			EnvVar: "RESERVE_ADDRESS_URL",
```
favadi: should be replaced with accounting/reserve-addresses/client.NewClientFlags.

favadi: ping?

halink0803: It is not convenient because we only need the endpoint url here, not the whole new client. If we use NewClientFlags, with NewClientFromContext function, we need to export `url` field or one more function to get client URL. It is more complicated.

favadi: fair point.

PR #267, opened 2019-03-27
2541_integrate_rsv_addr_to_fetcher_review_2

favadi:


reserverates/crawler/crawler.go
```diff
@@ -29,23 +29,16 @@ type ReserveRatesCrawler struct {
 // NewReserveRatesCrawler returns an instant of ReserveRatesCrawler.
 func NewReserveRatesCrawler(
 	sugar *zap.SugaredLogger,
-	addrs []string,
```
cyberliem: I tried not to change crawler's logic since currently reserve_rate_crawler of reserve_stats are running fine. This removal of address in crawler's construction will affect the package (already the test are failed). Should we do this in current scope or should we create a different task to refactor the reserveRateCrawler's logic?

favadi: > Should we do this in current scope or should we create a different task to refactor the reserveRateCrawler's logic?

we should do it here, not a too big changes as far as I see.

cyberliem: I changed the logic of rsv_rate crawler.

PR #270, opened 2019-03-27
2549 accounting build

vietthang207:


.travis.yml
```diff
@@ -34,10 +36,18 @@ before_script:
...
 
 deploy:
+  - provider: script
+    script: bash .travis/docker_build.sh
+    on:
+      tags: true
+      branch:
+        - develop
+        - staging
+        - master
```
favadi: why? the tests are running in docker_build.sh. We always need to test.

vietthang207: Sorry, I didn't read the scripts carefully. I've just renamed them to reflect their purposes. I guess that I don't need to do anything else regarding the script.

PR #287, opened 2019-04-02
2962 Update version to reserve addresses

halink0803:


accounting/reserve-addresses/storage/postgresql/storage.go
```diff
@@ -124,28 +155,33 @@ WHERE id = $1`
...
 // It returns no error if there is nothing in database.
-func (s *Storage) GetAll() ([]*common.ReserveAddress, error) {
+func (s *Storage) GetAll() ([]*common.ReserveAddress, int64, error) {
 	var (
 		logger    = s.sugar.With("func", "accounting/reserve-addresses/storage/postgresql/Storage.GetAll")
 		stored    []*ReserveAddress
 		results   []*common.ReserveAddress
 		queryStmt = `SELECT id, address, type, description, timestamp
 FROM addresses`
+		queryVersionStmt = fmt.Sprintf(`SELECT MAX(id) FROM %[1]s`, addressVersionTableName)
```
favadi: the id sequence might not be consecutive. For example, record 1 might have id = 2, record 2 might have id = 5. That's how sequence in PGSQL works. So, we should make id field normal INT field and do update manually: `UPDATE SET version = version + 1`.

PR #291, opened 2019-04-03
2970 Add test duplicate records

halink0803:


accounting/huobi/storage/postgres/postgres_test.go
```diff
@@ -77,4 +77,9 @@ func TestSaveAndGetAccountingRates(t *testing.T) {
 	require.NoError(t, err)
 	assert.Equal(t, len(data), 1)
 	assert.Equal(t, testData[0].FieldAmount, data[0].FieldAmount)
+
+	// test database does not stored duplicated records(with the same id)
+	data, err = hdb.GetTradeHistory(timeutil.TimestampMsToTime(1540793585679), timeutil.TimestampMsToTime(1540793586000))
```
cyberliem: without another call on Update(), obviously the db will remain the same

halink0803: Actually the first Update call on 11 records with 10 is the same id with different timestamp. I change the fromTime, toTime params. If it stored duplicate records, the Get function with return more than 1 record for that time range.

PR #294, opened 2019-04-04
2969 Make wallet and token address become optional

halink0803:


accounting/wallet-erc20/storage/postgres/postgres.go
```diff
@@ -149,14 +150,18 @@ func (wdb *WalletErc20Storage) GetERC20Transfers(wallet, token ethereum.Address,
...
 	)
-	const selectStmt = `SELECT data FROM %[1]s WHERE 
-	(data->>'from'=$1 OR data->>'to'=$1)
-	AND data->>'contractAddress'=$2
-	AND ((data->>'timestamp')>=$3::text AND (data->>'timestamp')<$4::text)
-	`
+	const selectStmt = `SELECT data FROM %[1]s WHERE ((data->>'timestamp')>=$1::text AND (data->>'timestamp')<$2::text)`
 	query := fmt.Sprintf(selectStmt, wdb.tableName)
+	if !blockchain.IsZeroAddress(wallet) {
+		query += fmt.Sprintf(` AND (data->>'from'='%[1]s' OR data->>'to'='%[1]s')`, wallet.Hex())
```
favadi: manually concat SQL string is risky. A trick to use with optional where condition is.

```SQL
SELECT * FROM users WHERE ($1 OR age > 18).
```

$1 (can be called: noFilterAge) type is boolean:

- TRUE --> the condition is ignored
- FALSE --> the condition is honored.

accounting/wallet-erc20/storage/postgres/postgres.go
```diff
@@ -149,14 +149,12 @@ func (wdb *WalletErc20Storage) GetERC20Transfers(wallet, token ethereum.Address,
 		)
 		tmp common.ERC20Transfer
 	)
-	const selectStmt = `SELECT data FROM %[1]s WHERE 
-	(data->>'from'=$1 OR data->>'to'=$1)
-	AND data->>'contractAddress'=$2
-	AND ((data->>'timestamp')>=$3::text AND (data->>'timestamp')<$4::text)
-	`
+	const selectStmt = `SELECT data FROM %[1]s WHERE ((data->>'timestamp')>=$1::text AND (data->>'timestamp')<$2::text) AND
+	($3 = '0x0000000000000000000000000000000000000000' OR (data->>'from'=$3 OR data->>'to'=$3)) AND
```
favadi: this duplicated `0x0000000000000000000000000000000000000000` does not look good, the comparison should be done in Go side using `IsZero` instead.

halink0803: changed using IsZeroAddress.

PR #313, opened 2019-04-11
show error message in huobi client error

favadi:


lib/huobi/huobi_client_test.go
```diff
@@ -14,7 +14,7 @@ import (
 )
 
 func TestHuobiClient(t *testing.T) {
-	t.Skip("skip huobi trade history test")
+	//t.Skip("skip huobi trade history test")
```
halink0803: This should be skip as external test.

favadi: fixed in 05247c5e6b94e0c5e13ce6b6fcb40b85ee486130.

PR #316, opened 2019-04-15
3103 Update version trigger and test

halink0803:


accounting/reserve-addresses/http/http_test.go
```diff
@@ -486,6 +486,32 @@ func TestVersion(t *testing.T) {
...
+			},
+		},
+		{
+			Msg:      "test version no changes",
+			Endpoint: "/addresses",
+			Method:   http.MethodGet,
+			Assert: func(t *testing.T, resp *httptest.ResponseRecorder) {
+				t.Helper()
+				require.Equal(t, http.StatusOK, resp.Code)
+
```
favadi: to be sure, could you add test cases when we try to INSERT/UPDATE but with invalid input (so the transaction rollbacked) and the version is not increased.

halink0803: In PUT request, there are 4 input: id, address, addressType and description. address, addressType are validated before sending to update, description are accepting all string (including empty), then there are only id can be invalid when processing in storage Update function. 

Therefore I added test with invalid id, which is not exists, assert function return Not Found and version not increase. Correct.

PR #317, opened 2019-04-16
3089 Add filter rerserve type

halink0803:


accounting/cmd/accounting-reserve-transaction-fetcher/main.go
```diff
@@ -75,7 +75,7 @@ func fetchTx(
 	// to hold any ERC20 tokens.
 	if addr.Type == common.Reserve || addr.Type == common.IntermediateOperator || addr.Type == common.CompanyWallet {
 		logger.Infow("fetching ERC20 transactions")
-		transfers, err := f.ERC20Transfer(addr.Address, fromBlock, toBlock)
+		transfers, err := f.ERC20Transfer(addr.Address, fromBlock, toBlock, addr.Type)
```
favadi: hijack the transfer data to add address type will lead to duplicated transaction.

For example:
- company wallet A transfer 100 KNC --> reserve address B
- there are 2 transfers: one with address-type=company_wallet, one without it.

To deal with this, I think we will need to create a table with address, type columns. A trade/transfer will be linked to one or more address.

This should be do for all transactions type, not just ERC20.

halink0803: Created new table with type.

accounting/cmd/accounting-reserve-transaction-fetcher/main.go
```diff
@@ -75,13 +75,13 @@ func fetchTx(
...
```
favadi: we don't need addr.Type anymore, right? As I see addr.Type is also passed to `s.StoreERC20Transfer`.

accounting/reserve-transaction-fetcher/storage/postgres/postgres.go
```diff
@@ -72,7 +74,8 @@ CREATE INDEX IF NOT EXISTS "%[2]s_time_idx" ON "%[2]s" ((data ->> 'timestamp'));
 -- create table tx erc20
 CREATE TABLE IF NOT EXISTS "%[3]s"
 (
-    data JSONB NOT NULL UNIQUE
+		data JSONB NOT NULL UNIQUE,
+		tx_hash text NOT NULL PRIMARY KEY
```
favadi: this seems wrong, as a single Ethereum transaction can have multiple ERC20 transfers.

accounting/reserve-transaction-fetcher/storage/postgres/postgres.go
```diff
@@ -113,18 +139,38 @@ func (s *Storage) TearDown() error {
...
 }
 
+//StoreReserve save fetching reserve address into database
+func (s *Storage) StoreReserve(reserve ethereum.Address, reserveType string) error {
+	var (
+		logger = s.sugar.With("func", "accounting/reserve-transaction-fetcher/storage/postgres/Storage.StoreReserve")
+	)
+	const storeReserve = `INSERT INTO "%[1]s" (address, address_type)
+	VALUES ($1, $2) 
+	ON CONFLICT (address) DO NOTHING`
```
favadi: the address type could be changed, we should update it on conflict.

accounting/reserve-transaction-fetcher/storage/postgres/postgres.go
```diff
@@ -22,6 +22,8 @@ var defaultTableNames = &tableNames{
 	Internal:     "rsv_tx_internal",
 	ERC20:        "rsv_tx_erc20",
 	LastInserted: "rsv_tx_last_inserted",
+	Reserves:     "rsv_tx_reserves",
+	TxsReserves:  "rsv_txs_reserves",
```
favadi: rsv_txs_reserves and rsv_tx_reserves is kind of confusing. Please think of another names.

accounting/reserve-transaction-fetcher/storage/postgres/postgres.go
```diff
@@ -82,8 +87,25 @@ CREATE INDEX IF NOT EXISTS "%[3]s_time_idx" ON "%[3]s" ((data ->> 'timestamp'),
 CREATE TABLE IF NOT EXISTS "%[4]s"
 (
     address       text   NOT NULL PRIMARY KEY,
-    last_inserted BIGINT NOT NULL
-);`
+		last_inserted BIGINT NOT NULL
+);
```
favadi: we should change the address column of this table to link to rsv_tx_reserves?

accounting/cmd/accounting-reserve-transaction-fetcher/main.go
```diff
@@ -81,7 +81,7 @@ func fetchTx(
 		}
 		logger.Infow("storing ERC20 transfers to database", "transfers", len(transfers))
 		if len(transfers) > 0 {
-			if err = s.StoreERC20Transfer(transfers); err != nil {
+			if err = s.StoreERC20Transfer(transfers, addr.Address); err != nil {
```
favadi: for consistent sake, we should do the same for internal/normal transactions.

accounting/reserve-transaction-fetcher/storage/postgres/postgres.go
```diff
@@ -72,18 +78,36 @@ CREATE INDEX IF NOT EXISTS "%[2]s_time_idx" ON "%[2]s" ((data ->> 'timestamp'));
...
+    address_key text UNIQUE REFERENCES "%[4]s" (address),
+		last_inserted BIGINT NOT NULL
+);
+
+-- create table link from tx to reserve
+CREATE TABLE IF NOT EXISTS "%[6]s"
+(
+	normal_tx_id int REFERENCES "%[1]s" (id),
+	internal_tx_id int REFERENCES "%[2]s" (id),
+	erc20_tx_id int REFERENCES "%[3]s" (id),
```
favadi: should have a constraint that ensure there is only one column not NULL of normal_tx_id/internal_tx_id/erc20_tx_id?

favadi: also, could you verify that if it is possible to insert duplicated data to this table (by fetching reserve-transaction with same from time / to time multiple times).

favadi: Also, please try if using 3 tables is easier than one with 3 columns for each type of tx.

accounting/reserve-transaction-fetcher/storage/postgres/postgres.go
```diff
@@ -260,19 +350,26 @@ WHERE data ->> 'timestamp' >= $1
...
 VALUES ($1)
-ON CONFLICT DO NOTHING;
+ON CONFLICT DO NOTHING RETURNING id;
 `
+	)
 
 	query := fmt.Sprintf(updateStmt, s.tableNames.ERC20)
 	logger.Debugw("storing ERC20 transfers to database", "query", query)
 
+	insertTxRsvQuery := s.storeTxsReservesQuery("erc20")
```
favadi: it does not look safe passing string value like that, maybe using multiple tables is better.

PR #320, opened 2019-04-19
2983  Replace custom table custom database

halink0803:


accounting/binance/storage/tradestorage/trade_storage.go
```diff
@@ -82,7 +59,7 @@ func (bd *BinanceStorage) Close() error {
 
 //DeleteTable remove trades table
 func (bd *BinanceStorage) DeleteTable() error {
```
favadi: as #322 is merged, all this DeleteTable, DeleteAllTables should be unused, right? Please double check.

PR #322, opened 2019-04-19
3146 Use different database for each service

halink0803:


accounting/common/common.go
```diff
@@ -2,5 +2,10 @@ package common
 
 // DefaultDB is the default database name of accounting service.
 const (
-	DefaultDB = "accounting"
+	DefaultCexTradesDB        = "cex_trades"
+	DefaultCexWithdrawalsDB   = "cex_withdrawals"
+	DefaultListedTokenDB      = "listed_tokens"
+	DefaultReserveRatesDB     = "reserve_rates"
+	DefaultTransactionsDB     = "transactions"
+	DefaultReserveAddressesDB = "reserve_addresses"
```
favadi: those database needs to be created in init-postgres-db.sh files.

halink0803: Added to init-postgres-db.sh

PR #325, opened 2019-04-22
custom unmarshalJSON for trade log

NNagato:


lib/timeutil/timeutil.go
```diff
@@ -23,3 +23,9 @@ func UnixMilliSecond() uint64 {
 func Midnight(t time.Time) time.Time {
 	return time.Date(t.Year(), t.Month(), t.Day(), 0, 0, 0, 0, t.Location())
 }
+
+// TimestampSecondToNs convert time stamp in second to nanosecond
+func TimestampSecondToNs(ts uint64) uint64 {
```
favadi: we use unix millis everywhere, please use that instead of nano for consistent sake.

tradelogs/common/types.go
```diff
@@ -77,6 +77,25 @@ func (tl *TradeLog) MarshalJSON() ([]byte, error) {
...
+	}
+	m := mask{
+		Timestamp:     0,
+		AliasTradeLog: (*AliasTradeLog)(tl),
+	}
+	err := json.Unmarshal(b, &m)
+	if err != nil {
+		return err
+	}
+	tl.Timestamp = time.Unix(0, int64(timeutil.TimestampSecondToNs(m.Timestamp)))
```
favadi: should be `tl.Timestamp = timeutil.TimestampMsToTime(m.Timestamp)`.

lib/timeutil/timeutil.go
NNagato: fixed in https://github.com/KyberNetwork/reserve-stats/pull/325/commits/ae988040fce8c3aeb4148d0f7d613b6608e769f7

tradelogs/common/types.go
NNagato: fixed in https://github.com/KyberNetwork/reserve-stats/pull/325/commits/147ee7e0e6dc1d09cd9149e742798832b60d5fc5

PR #326, opened 2019-04-23
3165 Update strategy retry for etherscan transaction fetcher

halink0803:


accounting/reserve-transaction-fetcher/fetcher/fetcher.go
```diff
@@ -32,6 +33,30 @@ func newFetchFunction(name string, fetch func(address string, startBlock *int, e
...
+	var (
+		txs    []interface{}
+		err    error
+		logger = f.sugar.With(
+			"func", "accounting-reserve-transaction/fetcher/fetchWithRetry",
+		)
+	)
+	for i := 0; i < f.attempt; i++ {
+		txs, err = fn.fetch(addr.String(), startBlock, endBlock, page, offset)
+		if blockchain.IsEtherscanTimeout(err) {
```
favadi: we should just retry every time got an error, the operator should tune the offset value if it still failed after retry.

favadi: also, please log the error != nil with a warning level message.

PR #328, opened 2019-04-24
3218 Add decimals to token info

halink0803:


lib/blockchain/token_info.go
```diff
@@ -31,9 +31,10 @@ var cachedName = map[common.Address]string{
 
 // TokenInfoGetter is a helper to get token info
 type TokenInfoGetter struct {
-	ethClient    bind.ContractBackend // eth client
-	cachedSymbol sync.Map
-	cachedName   sync.Map
+	ethClient      bind.ContractBackend // eth client
+	cachedSymbol   sync.Map
+	cachedName     sync.Map
+	cachedDecimals sync.Map
```
favadi: does it make sense / is it better/ possible to use a single map for symbol/name/decimals?

halink0803: No, I don't think it is better to use single map for both 3 fields. They are stored in map separately with same key "address". It is easier to get and cast type with 3 different map.

PR #330, opened 2019-04-24
improve fetch USD price

NNagato:


lib/tokenrate/cached_rate_provider.go
```diff
@@ -8,20 +8,42 @@ import (
...
-func NewCachedRateProvider(sugar *zap.SugaredLogger, provider tokenrate.ETHUSDRateProvider, timeout time.Duration) *CachedRateProvider {
-	return &CachedRateProvider{
+func NewCachedRateProvider(sugar *zap.SugaredLogger, provider tokenrate.ETHUSDRateProvider, options ...CachedRateProviderOption) *CachedRateProvider {
+	crp := &CachedRateProvider{
 		sugar:    sugar,
-		timeout:  timeout,
 		provider: provider,
 	}
+	for _, option := range options {
+		option(crp)
```
favadi: what will happen if user provides no timeout option? We should either handle it or provide a default value.

lib/tokenrate/cached_rate_provider.go
```diff
@@ -8,20 +8,42 @@ import (
...
 
 //CachedRateProvider is cached provider for eth usd rate
 type CachedRateProvider struct {
-	sugar    *zap.SugaredLogger
-	timeout  time.Duration
-	provider tokenrate.ETHUSDRateProvider
+	sugar       *zap.SugaredLogger
+	timeout     time.Duration
+	provider    tokenrate.ETHUSDRateProvider
+	withWarning bool
```
favadi: withWarning should be called warningOnly.

NNagato: fixed in https://github.com/KyberNetwork/reserve-stats/pull/330/commits/78ada81cd8816f253aef998174ebdf909b84d0c8

lib/tokenrate/cached_rate_provider.go
NNagato: I think we will provide a default value for timeout, fixed in https://github.com/KyberNetwork/reserve-stats/pull/330/commits/5976ca16a471a330477a99770db29f22e7543c39

PR #333, opened 2019-04-25
update reserve-stats broadcast client with authentication

NNagato:


lib/broadcast/broadcast_test.go
```diff
@@ -10,9 +10,14 @@ import (
...
-func newTestGeoInfo(server *httptest.Server) (*Client, error) {
+const (
+	readKeyID     = "readKeyID"
+	readSecretKey = "xxx123xxx"
+)
+
+func newTestGeoInfo(server *httptest.Server, readKeyID, readSecretKey string) (*Client, error) {
 	sugar := testutil.MustNewDevelopmentSugaredLogger()
-	return NewClient(sugar, server.URL)
+	return NewClient(sugar, server.URL, readKeyID, readSecretKey)
```
favadi: authentication should be optional, please create an option for NewClient constructor with name `WithAuth` or something.

lib/broadcast/flags.go
```diff
@@ -23,6 +26,16 @@ func NewCliFlags() []cli.Flag {
 			Value:  geoURLDefaultValue,
 			EnvVar: "GEO_URL",
 		},
+		cli.StringFlag{
+			Name:   readAccessKeyFlag,
+			Usage:  "key for access GET api",
+			EnvVar: "READ_ACCESS_KEY",
```
favadi: READ_ACCESS_KEY/READ_SECRET_KEY is too vague, should be something like GEO_ACCESS_KEY_ID, GEO_SECRET_ACCESS_KEY.

lib/broadcast/broadcast_test.go
```diff
@@ -90,3 +95,28 @@ func TestInvalidResponse(t *testing.T) {
...
+
+	g, err := newTestGeoInfo(server, "non-read-id", "non-read-secret")
+	if err != nil {
+		t.Error("Could not create Client object", "err", err.Error())
+	}
+	_, _, err = g.GetTxInfo(tx)
+	if err == nil {
+		t.Error("request must return error")
+	}
+}
```
favadi: This test really doesn't test anything, better to remove it.

lib/broadcast/flags.go
NNagato: fixed in https://github.com/KyberNetwork/reserve-stats/pull/333/commits/19c1cbee556a15987d45d632b10b1c8ccd07fbe3

lib/broadcast/broadcast_test.go
NNagato: fixed in https://github.com/KyberNetwork/reserve-stats/pull/333/commits/415839ef65a21757bdd9b4d4350a7275fd106560

lib/broadcast/broadcast_test.go
NNagato: fixed in https://github.com/KyberNetwork/reserve-stats/pull/333/commits/5ddd79553bee04de68e1ae3944101092bfdceefd

PR #335, opened 2019-04-25
3144 Import Huobi exported data

halink0803:


accounting/cmd/accounting-huobi-import-data/main.go
```diff
@@ -0,0 +1,256 @@
...
+				CreatedAt: updatedAt,
+			}
+			switch orderState {
+			case 5, 6:
+				order.FinishedAt = updatedAt
+			case 7:
+				order.CanceledAt = updatedAt
+			}
+			tradeHistories[orderID] = order
+		} else if orderState == 4 || (orderState == 6 && tradeHistories[orderID].State != "filled") {
```
favadi: 4, 5, and 6, 7 should be const, or better enum,  somewhere with descriptive name. I have no idea what they mean.

accounting/cmd/accounting-huobi-trade-fetcher/main.go
```diff
@@ -124,9 +125,37 @@ func run(c *cli.Context) error {
 			return err
 		}
 
-		var trades []huobi.TradeHistory
+		var trades = make(map[int64]huobi.TradeHistory)
 		for _, record := range data {
-			trades = append(trades, record...)
+			for _, trade := range record {
```
favadi: all logic inside should be moved to a function with comment about what it does.

halink0803: moved to another function.

PR #336, opened 2019-04-26
3259  Fix error insert blockhash to txhash

halink0803:


accounting/reserve-transaction-fetcher/storage/postgres/postgres.go
```diff
@@ -134,7 +134,7 @@ ON CONFLICT (tx_hash) DO UPDATE SET data = EXCLUDED.data RETURNING id;
 `
 		insertStmt = `INSERT INTO "rsv_tx_normal_tx_reserve" (tx_id, address_key)
 	VALUES ($1, $2)
-		ON CONFLICT DO NOTHING;`
+		ON CONFLICT (tx_id, address_key) DO NOTHING;`
```
favadi: could you explain the changes to ON CONFLICT statement?

halink0803: This is primary key, I just want to make sure it does not affect anything else.

favadi: I'm still not sure which issue does it fix. So please give a use-case when it is a problem or revert it.

halink0803: I want to prevent future issue might coming. It does not have use-case with problem right now. Reverted it.

PR #338, opened 2019-05-02
3233 Remove kyc information from users-stats

halink0803:


users/common/caps.go
```diff
@@ -29,36 +23,25 @@ type UserCap struct {
 // UserCapConfiguration is the cap configuration for KYC and non-KYC users.
 type UserCapConfiguration struct {
 	NonKYC UserCap
-	KYC    UserCap
 }
 
 // NewUserCapConfiguration creates new instance of UserCapConfiguration from given parameters.
-func NewUserCapConfiguration(nonKYCDailyLimit, nonKYCTxLimit, kycDailyLimit, kycTxLimit float64) *UserCapConfiguration {
```
favadi: we still need the kyc cap for users stats by uid API.

PR #349, opened 2019-05-07
3285 push eth sent transfer events to chainalysis api

NNagato:


tradelogs/chainalysis/client.go
```diff
@@ -0,0 +1,171 @@
...
+	req.Header.Add("Token", c.apiKey)
+	resp, err := c.client.Do(req)
+	if err != nil {
+		return err
+	}
+	defer func() {
+		if cErr := resp.Body.Close(); cErr != nil {
+			c.sugar.Errorw("failed to close body", "err", cErr.Error())
+		}
+	}()
```
favadi: we need to verify status code of the request and display its response body for debug purpose.

tradelogs/chainalysis/client.go
```diff
@@ -0,0 +1,171 @@
...
+		if strings.ToLower(log.DestAddress.Hex()) != ethAddress {
+			continue
+		}
+
+		c.sugar.Debugw("sent transfer data",
+			"user addr", userAddress,
+			"receive addr", receiveAdderss,
+			"tx hash", txHash)
+		if rd, ok := mapRegisterData[userAddress]; ok {
+			mapRegisterData[userAddress] = updateRegisterData(rd, ethSymbol, txHash, receiveAdderss)
```
favadi: I don't see the point of storing data in a middle map and doing the requests later. What is its advantage over just call 2 requests here?

tradelogs/cmd/trade-logs-chainalysis/main.go
```diff
@@ -0,0 +1,88 @@
...
+
+	libapp "github.com/KyberNetwork/reserve-stats/lib/app"
+	"github.com/KyberNetwork/reserve-stats/lib/timeutil"
+	"github.com/KyberNetwork/reserve-stats/tradelogs/chainalysis"
+	"github.com/KyberNetwork/reserve-stats/tradelogs/client"
+)
+
+const (
+	readTradeLogAccessKeyIDFlag     = "trade-log-read-access-key-id"
+	readTradeLogSecretAccessKeyFlag = "trade-logread-secret-access-key"
```
favadi: "trade-logread-secret-access-key" doesn't look right.

favadi: also, no need to indicate that they are "read" key.

tradelogs/cmd/trade-logs-chainalysis/main.go
```diff
@@ -0,0 +1,88 @@
...
+		return err
+	}
+
+	opt := client.WithAuth(c.String(readTradeLogAccessKeyIDFlag), c.String(readTradeLogSecretAccessKeyFlag))
+	tradeLogCli, err := client.NewClientFromContext(sugar, c, opt)
+	if err != nil {
+		return err
+	}
+
+	tradeLogs, err := tradeLogCli.GetTradeLogs(fromTime, toTime)
```
favadi: that won't work if the time window is too large. Could you check max allowed time range of get trades logs API? We might need to break to smaller chunk and call the API multiple times.

tradelogs/cmd/trade-logs-chainalysis/main.go
```diff
@@ -0,0 +1,88 @@
...
+	if err != nil {
+		return err
+	}
+
+	chainAlysisCli, err := chainalysis.NewClientFromContext(sugar, c)
+	if err != nil {
+		return err
+	}
+
+	return chainAlysisCli.PushETHSentTransferEvent(tradeLogs)
```
favadi: we need to add retry function, in case of network error.

tradelogs/common/types.go
```diff
@@ -38,6 +38,7 @@ type TradeLog struct {
 	DestSymbol  string           `json:"dst_symbol,omitempty"`
 
 	UserAddress       ethereum.Address `json:"user_addr"`
+	ReceiveAddress    ethereum.Address `json:"receive_address"`
```
favadi: should be call ReceiverAddress (user <-> receiver).

PR #351, opened 2019-05-10
3329  Detect if a transfer erc20 transaction is belong to a trade or not

halink0803:


accounting/trades-detect/detect.go
```diff
@@ -0,0 +1,33 @@
...
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+	receipt, err := ethClient.TransactionReceipt(ctx, txHash)
+	if err != nil {
+		return false, err
+	}
+	for _, log := range receipt.Logs {
+		for _, topic := range log.Topics {
+			if topic == ethereum.HexToHash(tradeExecuteEvent) {
+				return true, nil
```
favadi: the logic is not good enough, (in theory) there will be case when user do both:

- ERC20 transfer token T1 to reserve R1 (1)
- kyber trade token T2 --> ETH
  - ERC20 transfer T2: user --> kyber network proxy (2)
  - ERC20 transfer T2: kyber network proxy --> reserve R1 (3)

So in a transaction, (1) should not be marked as is_trade.

halink0803: I don't know if (1) does not belong to any trade then its receipt should not have ExecuteTrade event, right? In this situation, how could I check the tx is belong to a trade?

favadi: > I don't know if (1) does not belong to any trade then its receipt should not have ExecuteTrade event, right? In this situation, how could I check the tx is belong to a trade?

(1) is still in the same transaction, so it is included in the transaction receipt. In above case, we can track all the transactions that send ERC20 token to reserve right before the tradeExecuteEvent. This way we know that (1) is not belong the trade.

PR #355, opened 2019-05-14
3335 Optimize users internal api

halink0803:


users/cacher/internal_cacher.go
```diff
@@ -0,0 +1,107 @@
...
+	return nil
+}
+
+func (irc *InternalRedisCacher) cache24hVolumeByUID() error {
+	var (
+		logger = irc.sugar.With("func", "user/cacher/cache24hVolumeByUID")
+	)
+
+	// read total trade 24h
+	query := fmt.Sprintf(`SELECT SUM(amount) FROM (SELECT %s*%s as amount FROM trades WHERE time >= (now()-24h)) GROUP BY uid`,
```
favadi: should we skip all records with uid == NULL?

PR #357, opened 2019-05-16
3384 Optimize is trade detection function

halink0803:


accounting/reserve-transaction-fetcher/fetcher/fetcher.go
```diff
@@ -150,6 +156,12 @@ func (f *EtherscanTransactionFetcher) NormalTx(addr ethereum.Address, from, to *
 
 // InternalTx returns all internal transaction of given address between block range.
 func (f *EtherscanTransactionFetcher) InternalTx(addr ethereum.Address, from, to *big.Int, offset int) ([]common.InternalTx, error) {
+	var (
+		logger   = f.sugar.With("func", "reserve-transaction-fetcher/InternalTx")
+		temp     sync.Map
```
favadi: temp is bad variable name.

PR #362, opened 2019-06-04
3650 add tx sender to tradelogs

ductm54:


tradelogs/common/types.go
```diff
@@ -54,9 +54,10 @@ type TradeLog struct {
 	WalletFees     []WalletFee `json:"-"`
 	IntegrationApp string      `json:"integration_app"`
 
-	IP      string `json:"ip"`
-	Country string `json:"country"`
-	UID     string `json:"uid"`
+	IP      string           `json:"ip"`
+	Country string           `json:"country"`
+	UID     string           `json:"uid"`
+	Sender  ethereum.Address `json:"tx_sender"`
```
favadi: json symbol should match variable name.

tradelogs/crawler_v3.go
```diff
@@ -50,6 +50,11 @@ func (crawler *Crawler) assembleTradeLogsV3(eventLogs []types.Log) ([]common.Tra
 			return result, errors.New("log item has no topic")
 		}
 
+		tradeLog.Sender, err = crawler.getTxSender(log, 10*time.Second)
```
favadi: so each log will require one additional request to ethereum node, which will add significant processing time to the crawler. Could you confirm that there is nothing in our current event logs that provider tx sender?

favadi: Also, 

- 10*time.Second could be a const somewhere.
- there might be multiple trades in a single tx hash, and we might unnecessary get tx sender of a transaction multiple times. Does it make sense to store the map tx_hash --> sender to reduce API calls?

tradelogs/common/types.go
ductm54: fixed

tradelogs/crawler_v3.go
ductm54: - Our current event logs does not provide tx sender.
- change 10*time.Second to default timeout
- I think we might not need map tx_hash -> sender as most of time we will not hit cache

PR #372, opened 2019-07-22
add volume * 2 when swap token to token

ducquangkstn:


tradelogs/crawler_v1.go
```diff
@@ -83,6 +84,11 @@ func (crawler *Crawler) getInternalTransaction(tradeLog common.TradeLog) (common
 	blockInt := int(tradeLog.BlockNumber)
 	internalTxs, err := crawler.etherscanClient.InternalTxByAddress(internalNetworkAddrV1, &blockInt, &blockInt, 0, 0, false)
 	if err != nil {
+		if strings.HasPrefix(err.Error(), "etherscan server:") {
```
halink0803: This does not look nice, use this: https://github.com/KyberNetwork/reserve-stats/blob/3cd18e220de5d400d1f50911cd3ce87768c038e7/lib/blockchain/contract_timestamp_etherscan.go#L13

tradelogs/crawler_v2.go
```diff
@@ -43,29 +43,27 @@ func (crawler *Crawler) fetchTradeLogV2(fromBlock, toBlock *big.Int, timeout tim
...
-				break
+				if !blockchain.IsZeroAddress(log.Address) {
+					return log.Address
+				}
 			}
 		}
-		if !blockchain.IsZeroAddress(reserveAddr) {
-			break
-		}
 	}
```
halink0803: I think we should return zero address here. What happen if the for go through and no topic equal tradeExecuteEvent?

PR #380, opened 2019-08-09
fix cap returned = min(tx_limit, daily_limit - volume)

ducquangkstn:


users/http/server.go
```diff
@@ -184,18 +184,19 @@ func (s *Server) userStatsBatch(c *gin.Context) {
...
 		jsonOutput = append(jsonOutput, gin.H{
-			"cap":    userCap,
-			"kyced":  kycedArr[i],
-			"rich":   rich,
-			"volume": volumeInWei,
+			"cap":       userCap,
+			"kyced":     kycedArr[i],
+			"rich":      rich,
+			"volume":    volumeInWei,
+			"cap_daily": availableUserCapDaily,
```
halink0803: This should be remaining_daily_cap for more clear. please change the log also.

PR #383, opened 2019-09-13
add field trade_volume

ducquangkstn:


secmask:
look like TradeVolume is an other name of `EthAmount`, why we need it?

ducquangkstn:
> look like `TradeVolume` is an other name of `EthAmount`, why we need it?

Since the latest version, eth_amount is double for burnable token trade. But there are few reports that need the old version of eth_amount, so trade_volume is added for querying

ducquangkstn:
Rename field to original_eth_amount to clarify the purpose of this field

PR #384, opened 2019-09-24
update logger with dynamic function name

ducquangkstn:


ducquangkstn:
This is just refactor PR, but it changes a lot of files. Reviewing this task should be the low priority

lib/caller/function_test.go
```diff
@@ -0,0 +1,29 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestGetCallerFunctionName(t *testing.T) {
+	names := firstFunction()
+	require.Equal(t, len(names), 3)
+	assert.Equal(t, "github.com/KyberNetwork/reserve-stats/lib/caller.firstFunction", names[0])
```
secmask: I would recommend to strip `{domain}/{organize}/` from function name, we just need to keep `reserve-stats/lib/caller.firstFunction`, so we can make a copy of `GetCurrentFunctionName` like `GetCurrentFunctionNameShort(2)` which will strip 2 first components in func name.

secmask:
@halink0803 this may affect if we search func name in Kibana

lib/caller/function_test.go
ducquangkstn: I have added a function to remove 2 first components in func name if func name has more than 2 components

PR #402, opened 2019-11-13
5012 create service store eth price from coinbase

NNagato:


tokenprice/cmd/token-price-api/main.go
```diff
@@ -0,0 +1,52 @@
...
+
+	"github.com/urfave/cli"
+
+	libapp "github.com/KyberNetwork/reserve-stats/lib/app"
+	"github.com/KyberNetwork/reserve-stats/tokenprice/server"
+	"github.com/KyberNetwork/reserve-stats/tokenprice/storage"
+)
+
+const (
+	hostFlag = "host"
```
secmask: `host` is not correct term for this, should be call `bindAddress`

tokenprice/cmd/token-price-api/main.go
```diff
@@ -0,0 +1,52 @@
...
+	app.Name = "Token rate crawler"
+	app.Usage = "Crawl token rate from some EX"
+	app.Version = "0.0.1"
+	app.Action = run
+
+	app.Flags = append(app.Flags,
+		cli.StringFlag{
+			Name:   hostFlag,
+			Usage:  "provide host for token price api",
+			EnvVar: "HOST",
```
secmask: should let a default value here

tokenprice/cmd/token-price-crawler/main.go
```diff
@@ -0,0 +1,174 @@
...
+
+const (
+	fromTimeFlag = "from-time"
+	toTimeFlag   = "to-time"
+)
+
+func main() {
+	app := libapp.NewApp()
+	app.Name = "Token rate crawler"
+	app.Usage = "Crawl token rate from some EX"
```
secmask: need a better usage help

tokenprice/cmd/token-price-crawler/main.go
```diff
@@ -0,0 +1,174 @@
...
+	app.Version = "0.0.1"
+	app.Action = run
+
+	app.Flags = append(app.Flags,
+		cli.StringFlag{
+			Name:   fromTimeFlag,
+			Usage:  "provide from time to crawl token rate",
```
secmask: need an example of `from-time` - `to-time` value, I dont know how to set it.

tokenprice/cmd/token-price-crawler/main.go
```diff
@@ -0,0 +1,174 @@
...
+}
+
+func run(c *cli.Context) error {
+	sugar, flush, err := libapp.NewSugaredLogger(c)
+	if err != nil {
+		return err
+	}
+	defer flush()
+
+	p, err := provider.NewProvider(provider.Coinbase)
```
secmask: what will provider provide? we need a better name

tokenprice/cmd/token-price-crawler/main.go
```diff
@@ -0,0 +1,174 @@
...
+		logger.Info("from-time and to-time are blank, run get token rate daily...")
+		if err := crawlTokenRateDaily(sugar, ethID, usdID, p, s); err != nil {
+			logger.Errorw("failed to crawl token rate daily", "error", err)
+			return err
+		}
+	}
+
+	logger = logger.With("from time", fromTimeS, "to time", toTimeS)
+
+	fromeTime, toTime, err := validateTime(fromTimeS, toTimeS)
```
secmask: please extract crawl with time-range into function

tokenprice/cmd/token-price-crawler/main.go
```diff
@@ -0,0 +1,174 @@
...
+		if err := s.SaveTokenRate(token, currency, p.Name(), now, rate); err != nil && err != postgres.ErrExists {
+			logger.Errorw("failed to save data to database", "error", err)
+			errCh <- err
+		}
+		logger.Info("save token rate successfully")
+	}
+	// get rate today
+	job()
+
+	if _, err := scheduler.Every().Day().At("00:00:01").Run(job); err != nil {
```
secmask: we need to run this job at UTC time, not local time.

tokenprice/storage/postgres/postgres.go
```diff
@@ -0,0 +1,116 @@
...
+// TokenRateDB is storage of token rate
+type TokenRateDB struct {
+	sugar *zap.SugaredLogger
+	db    *sqlx.DB
+}
+
+// NewTokenRateDB return instance of TokenRateDB
+func NewTokenRateDB(sugar *zap.SugaredLogger, db *sqlx.DB) (*TokenRateDB, error) {
+	const (
+		tokenRateSchema = `
```
secmask: should move init table stuff to separate function as we may have migration also later.

tokenprice/storage/postgres/postgres.go
```diff
@@ -0,0 +1,116 @@
...
+		VALUES (DATE($1), $2, $3, $4, $5)`
+	)
+	logger.Infow("save new token rate", "query", query)
+	log.Println(query)
+	_, err := trdb.db.Exec(query, timestamp, source, token, currency, rate)
+	if err != nil {
+		// check if return error is a known pq error
+		pErr, ok := err.(*pq.Error)
+		if !ok {
+			return err
```
secmask: wrap with `errors` so we can trace it later.

tokenprice/storage/postgres/postgres.go
```diff
@@ -0,0 +1,116 @@
...
+		}
+
+		logger.Infow("got error from database",
```
secmask: why info an error?

tokenprice/storage/postgres/postgres.go
```diff
@@ -0,0 +1,116 @@
...
+		query = `SELECT value FROM "tokenrates" 
+			WHERE token=$1 AND currency=$2 AND date=DATE($3)`
+
+		dbResult tokenRateDB
+	)
+	logger.Infow("get token rate", "query", query)
+	if err := trdb.db.Get(&dbResult, query, token, currency, timestamp); err == sql.ErrNoRows {
+		return 0, ErrNotFound
+	} else if err != nil {
+		logger.Errorw("got error from database", "error", err.Error())
```
secmask: `"error", err.Error())` -> just `err` is ok, we dont need to call `Error()`

tokenprice/storage/postgres/postgres.go
```diff
@@ -0,0 +1,116 @@
...
+		return 0, errors.New("failed to query token rate in database")
+	}
+	if !dbResult.Rate.Valid {
+		return 0, errors.New("invalid data return")
```
secmask: when a rate can be invalid?

tokenprice/cmd/token-price-crawler/main.go
```diff
@@ -0,0 +1,174 @@
...
+		logger = sugar.With("func", caller.GetCurrentFunctionName(),
+			"token", token, "currency", currency, "provider", p.Name())
+	)
+	job := func() {
+		logger.Info("Running job")
+		var now = time.Now().UTC()
+		rate, err := p.Rate(token, currency, now)
+		if err != nil {
+			logger.Errorw("failed to get token rate", "error", err)
+			errCh <- err
```
secmask: why dont we just return `err` instead of put it into chan?

NNagato: I use lib https://godoc.org/github.com/carlescere/scheduler#Job.Run to create a cron job, and the input require a function without return, so i have to  put the error to channel

tokenprice/storage/postgres/postgres.go
NNagato: changed as request

tokenprice/storage/postgres/postgres.go
NNagato: changed as request

tokenprice/cmd/token-price-crawler/main.go
NNagato: changed as request

tokenprice/cmd/token-price-crawler/main.go
secmask: there's handling err problem here, err can be generate by job() in future, and then no one consume the errChan.

tokenprice/cmd/token-price-crawler/main.go
```diff
@@ -0,0 +1,187 @@
...
+	)
+
+	app.Flags = append(app.Flags, libapp.NewPostgreSQLFlags(storage.DefaultDB)...)
+	app.Flags = append(app.Flags, libapp.NewSentryFlags()...)
+	if err := app.Run(os.Args); err != nil {
+		log.Fatal(err)
+	}
+}
+
+func validateTime(fromTimeS, toTimeS string) (time.Time, time.Time, error) {
```
secmask: this validate allow both `fromTimeS, toTimeS` empty is valid?

tokenprice/cmd/token-price-crawler/main.go
```diff
@@ -0,0 +1,187 @@
...
+	<-cs
+	logger.Info("got interrupt signal, program exited")
+	return nil
+}
+
+func crawlTokenPriceWithTimeRange(sugar *zap.SugaredLogger, fromTime, toTime time.Time, p provider.PriceProvider, s storage.Storage) error {
+	var (
+		logger = sugar.With("func", caller.GetCurrentFunctionName())
+
+		timeW8PerRequest = 1 * time.Second
```
secmask: this should be configurable base on provider, some provider has limit, some not.

tokenprice/cmd/token-price-crawler/main.go
```diff
@@ -0,0 +1,187 @@
...
+		}
+		logger.Info("save token price successfully")
+	}
+	// get price today
+	job()
+
+	// run job get price daily
+	l, _ := time.LoadLocation("Local")
+	firstMomentOfDay := time.Now().Truncate(24 * time.Hour).In(l).Format("15:04:05")
+	if _, err := scheduler.Every().Day().At(firstMomentOfDay).Run(job); err != nil {
```
secmask: `firstMomentOfDay` should be configurable instead of fixed time point.

tokenprice/cmd/token-price-crawler/main.go
secmask: `source` and `provider` are 2 term about same thing, we should use 1 to keep it consistent.

tokenprice/provider/coingecko/flags.go
```diff
@@ -0,0 +1,29 @@
...
+	"github.com/urfave/cli"
+)
+
+const (
+	reqTimeWaitingFlag    = "coingecko-req-waiting-time"
+	defaultReqTimeWaiting = time.Second
+)
+
+// NewFlags return cli config for coingecko
+func NewFlags() []cli.Flag {
```
secmask: this flag not use anywhere?

tokenprice/provider/coinbase/coinbase.go
```diff
@@ -0,0 +1,114 @@
...
+// CoinBase is the CoinBase implementation of Provider. The
+// precision of CoinBase provider is up to day.
+type CoinBase struct {
+	client         *http.Client
+	baseURL        string
+	reqWaitingTime time.Duration
+}
+
+// New creates a new CoinBase instance.
+func New(reqWaitingTime time.Duration) *CoinBase {
```
secmask: `reqWaitingTime` has no use?

PR #408, opened 2019-11-25
fix CachedRateProvider not depends on timestamp

ducquangkstn:


lib/tokenrate/cached_rate_provider.go
```diff
@@ -40,68 +32,38 @@ func NewCachedRateProvider(sugar *zap.SugaredLogger, provider tokenrate.ETHUSDRa
...
-
-		logger.Debugw("got ETH/USD rate",
-			"rate", rate)
-		crp.mu.Lock()
-		crp.cachedRate = rate
-		crp.cachedTime = time.Now()
-		crp.mu.Unlock()
-		return rate, nil
+		logger.Debugw("got ETH/USD rate", "rate", rate)
+		crp.cache.Set(timestamp.Format(time.RFC3339), rate, crp.timeout)
```
secmask: because today rate is not finally yet, so it will have shorter cache duration.

lib/tokenrate/cached_rate_provider.go
```diff
@@ -40,68 +32,38 @@ func NewCachedRateProvider(sugar *zap.SugaredLogger, provider tokenrate.ETHUSDRa
...
+			"timestamp", timestamp,
+		)
 	)
-
-	crp.mu.RLock()
-	cachedRate := crp.cachedRate
-	cachedTime := crp.cachedTime
-	crp.mu.RUnlock()
-	if cachedRate == 0 || cachedTime.IsZero() || crp.isCacheExpired() {
+	if item, found := crp.cache.Get(timestamp.Format(time.RFC3339)); !found {
```
secmask: I think we should truncate the key into format `yyyy-mm-dd`, string in `RFC3339` format will require caller pass an already truncated time(and same Location) to match the cached entry.

lib/tokenrate/cached_rate_provider.go
secmask: I would suggest 5 min for today rate.

ducquangkstn: change as request

lib/tokenrate/cached_rate_provider.go
ducquangkstn: change as request

PR #410, opened 2019-11-29
Api for report bot

halink0803:


tradelogs/reserve_names.go
```diff
@@ -0,0 +1,62 @@
...
+	ethereum.HexToAddress("0xcf1394c5e2e879969fdb1f464ce1487147863dcb"): "Oasis bridge reserve - v2",
+	ethereum.HexToAddress("0xAA14DCAA0AdbE79cBF00edC6cC4ED17ed39240AC"): "DAO stack APR reserve",
+	ethereum.HexToAddress("0xb45C8956a080d336934cEE52A35D4dbABF025b6F"): "Maker APR reserve",
+	ethereum.HexToAddress("0x05461124c86c0ad7c5d8e012e1499fd9109ffb7d"): "Gnosis APR reserve",
+	ethereum.HexToAddress("0x4Cb01bd05E4652CbB9F312aE604f4549D2bf2C99"): "Synth USD APR reserve",
+}
+
+// ReserveAddressToName return reserve name by its address
+func ReserveAddressToName(address ethereum.Address) string {
+	return reserves[address]
```
secmask: we got several trouble with default "" when access not exist map entry, so it's better to explicit return error when not found, and let caller decide what to do.

PR #440, opened 2020-01-08
Update check big trade

halink0803:


tradelogs/storage/postgres/big_trades.go
```diff
@@ -0,0 +1,106 @@
...
+	return result, nil
+}
+
+// SaveBigTrades save trades into db
+func (tldb *TradeLogDB) SaveBigTrades(bigVolume float32, fromBlock uint64) error {
+	var (
+		logger    = tldb.sugar.With("func", caller.GetCurrentFunctionName())
+		bigTrades = []uint64{}
+	)
+	if err := tldb.db.Select(&bigTrades, detectBigTradeQuery, bigVolume, fromBlock); err != nil {
```
secmask: note that select & insert here can be merge into single query, but there's not such many big trades so it does not help improve performance much.

tradelogs/cmd/trade-logs-crawler/main.go
```diff
@@ -44,6 +44,9 @@ const (
 
 	blockConfirmationsFlag    = "wait-for-confirmations"
 	defaultBlockConfirmations = 7
+
+	bigVolumeFlag    = "big-volume"
```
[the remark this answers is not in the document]
halink0803: Updated

tradelogs/storage/postgres/big_trades.go
halink0803: Yep, it does not improve performance much, but I think it would make the code look more clean, then I decided to change it that way.

PR #465, opened 2020-04-02
resolve tradeLog.WalletAddress from tx on no WalletFee trade

secmask:
- replace use WalletFee with tradeLog.WalletAddress in approxiate case
- retry to resolve tradeLog.WalletAddress from tx on no WalletFee trade

### DIFF
--- a/tradelogs/cmd/trade-logs-crawler/main.go
+++ b/tradelogs/cmd/trade-logs-crawler/main.go
@@ -7,7 +7,6 @@ import (
 	"os"
 	"time"
 
-	common2 "github.com/ethereum/go-ethereum/common"
 	"github.com/influxdata/influxdb/client/v2"
 	"github.com/urfave/cli"
 	"go.uber.org/zap"
@@ -15,6 +14,7 @@ import (
 	libapp "github.com/KyberNetwork/reserve-stats/lib/app"
 	"github.com/KyberNetwork/reserve-stats/lib/blockchain"
 	"github.com/KyberNetwork/reserve-stats/lib/broadcast"
+	"github.com/KyberNetwork/reserve-stats/lib/contracts"
 	"github.com/KyberNetwork/reserve-stats/lib/cq"
 	"github.com/KyberNetwork/reserve-stats/lib/deployment"
 	"github.com/KyberNetwork/reserve-stats/lib/etherscan"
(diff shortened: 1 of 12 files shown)


tradelogs/crawler_v3.go
```diff
@@ -74,6 +120,26 @@ func (crawler *Crawler) assembleTradeLogsV3(eventLogs []types.Log) ([]common.Tra
 			if err != nil {
 				return nil, errors.Wrapf(err, "failed to get transaction receipt tx: %v", tradeLog.TransactionHash)
 			}
+			if len(tradeLog.WalletFees) == 0 { // in case there's no fee, we try to get wallet addr from tradeWithHint input
```
ductm54, line 123: pls consider merge this one in [updateBasicInfo](https://github.com/KyberNetwork/reserve-stats/blob/develop/tradelogs/crawler.go#L366) and remove set wallet name and address in crawler v1 and v2
As in updateBasicInfo already call getTransactionByHash

tradelogs/crawler_v3.go
```diff
@@ -74,6 +120,26 @@ func (crawler *Crawler) assembleTradeLogsV3(eventLogs []types.Log) ([]common.Tra
...
+				tx, err := crawler.getTransactionByHash(tradeLog.TransactionHash, defaultTimeout)
+				if err != nil {
+					return nil, errors.Wrapf(err, "failed to get transactionByHash tx: %v", tradeLog.TransactionHash)
+				}
+				if bytes.Equal(tx.To().Bytes(), crawler.networkProxy.Bytes()) { // try to fail early, tx must have dst == networkProxy
+					tradeParam, err := decodeTradeWithHintParam(tx.Data())
+					if err == nil {
```
ductm54, line 130: should we handle edge case 1st

tradelogs/crawler_v2.go
```diff
@@ -53,6 +53,16 @@ func (crawler *Crawler) getTransactionReceipt(txHash ethereum.Hash, timeout time
 	return receipt, nil
 }
 
+func (crawler *Crawler) getTransactionByHash(txHash ethereum.Hash, timeout time.Duration) (*types.Transaction, error) {
```
ductm54, line 56: this one does not  use anywhere

tradelogs/cmd/trade-logs-crawler/main.go
```diff
@@ -215,15 +225,14 @@ func run(c *cli.Context) error {
 	if err != nil {
 		return err
 	}
-
+	networkProxyAddr := common2.HexToAddress(c.String(networkProxyAddrFlag))
```
halink0803, line 228: I think we already have the proxyContracts address here at: https://github.com/KyberNetwork/reserve-stats/blob/8fe023b2a259b0534cc15301641f23dbe17111a2/lib/contracts/addresses.go#L70.

Flag bring the flexible but I think it is redundant here.

PR #482, opened 2020-04-13
Update binance trade storage

halink0803:


accounting/binance/fetcher/fetcher.go
```diff
@@ -76,19 +78,11 @@ func (f *Fetcher) getTradeHistoryForOneSymBol(fromID uint64, symbol string) ([]b
 }
 
 //GetTradeHistory get all trade history from trades for all token
```
secmask, line 80: please update this comment docs as now this function do fetch and save data too.

halink0803: updated the comment

PR #510, opened 2020-09-25
update build with go 1.14

secmask:


halink0803 (approved):
LGTM

secmask:
I disable `unsed` check in test file due it has a bug, see https://github.com/golangci/golangci-lint/issues/791

PR #530, opened 2020-11-27
Update binance trade fetcher

halink0803:


lib/coreclient/client.go
```diff
@@ -0,0 +1,140 @@
...
+	if err != nil {
+		return respBody, err
+	}
+	defer func() {
+		if cErr := resp.Body.Close(); cErr != nil {
+			logger.Errorw("Response body close error", "error", cErr.Error())
+		}
+	}()
+	switch resp.StatusCode {
+	case 200:
```
secmask, line 95: use http.StatusOK

lib/coreclient/client.go
```diff
@@ -0,0 +1,140 @@
...
+		respBody, err = ioutil.ReadAll(resp.Body)
```
secmask, line 96: consider to move ReadAll out of switch block as we read body on all case

accounting/cmd/accounting-binance-frequent-trade-fetcher/main.go
```diff
@@ -0,0 +1,159 @@
...
+		}
+
+		binanceFetcher := fetcher.NewFetcher(sugar, binanceClient, retryDelay, attempt, batchSize, binanceStorage, account.Name, marketDataClient)
+
+		if err := binanceFetcher.GetTradeHistory(fromIDs, tokenPairs, account.Name); err != nil {
+			return err
+		}
+	}
+
+	return binanceStorage.Close()
```
secmask, line 158: should we use defer for Close here?

halink0803: just realized that we already have defer func for storage close.

halink0803: https://github.com/KyberNetwork/reserve-stats/blob/3babee9367424fca17cb51f65e8dc70087515d92/accounting/cmd/accounting-binance-frequent-trade-fetcher/main.go#L101

secmask (approved):

PR #607, opened 2022-02-23
update pricing good for left over on-chain trades

halink0803:


accounting/cex-trade/http/trades.go
```diff
@@ -454,84 +454,98 @@ func quoteFromOriginalSymbol(re *regexp.Regexp, symbol string) string {
...
+		if lastOnchainTrades.Type == buyType {
+			pnlRate = rebalanceAVGPrice / onchainAVGPrice
+		} else {
+			pnlRate = onchainAVGPrice / rebalanceAVGPrice
+		}
+		s.sugar.Infow("rate", "pnlRate", pnlRate)
+	} else {
+		pnlRate = 0
+	}
+	if pnlRate > 1.0010 {
```
secmask: let's define a const as I see it use in some place.

secmask (approved):

